PR #17, opened 2018-11-06
[AV2-23]activity detail component

shawnm0705:


trtshen:
@shawnm0705 Looks good! thanks

PR #19, opened 2018-11-07
Bugfix/av2 48/chat module routing issue

sasangachathumal:


src/app/chat/chat-room/chat-room.component.html
```diff
@@ -1,72 +1,35 @@
...
+            <div class="seen-text">
+              <p>{{message.sender_name}}</p>
+            </div>
+          </ng-container>
+        </ion-label>
+      </ion-item>
+    </ng-container>
   </ion-list>
 </ion-content>
 <ion-footer>
```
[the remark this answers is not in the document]
shawnm0705: and a typo `you'r`

src/app/chat/chat.component.html
```diff
@@ -1,69 +1,31 @@
...
+        <ion-label>
+          <div class="chat-name-container">
+            <h2>{{chat.is_team ? chat.team_name : chat.name}}</h2>
+            <ng-container *ngIf="chat.role === 'mentor'"><span>mentor - {{chat.team_name}}</span></ng-container>
+          </div>
+          <ng-container *ngIf="chat.last_message; else messgeElseBlock"><p>{{chat.last_message}}</p></ng-container>
+          <ng-template #messgeElseBlock><p>No messages.</p></ng-template>
+        </ion-label>
+        <div slot="end" class="time-container">
+          <span>yesterday</span>
```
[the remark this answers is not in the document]
sasangachathumal: variables for what. for else block?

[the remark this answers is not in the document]
shawnm0705: `yesterday` is hardcoded

src/app/chat/chat-routing.module.ts
```diff
@@ -7,10 +7,11 @@ import { ChatRoomComponent } from './chat-room/chat-room.component';
 const routes: Routes = [
   {
     path: '',
-    component: ChatComponent
+    redirectTo: '/chat/chat-room',
+    pathMatch: 'full'
   },
   {
-    path: 'chatroom',
+    path: 'chat-room',
```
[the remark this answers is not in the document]
sasangachathumal: if we include id that means we can call api using that id and get selected chat data or we can search it on our cashed data. To get one chat object no need to make api call, and also we didn't cash chat data because it's a real time function. We can pass selected chat data from chat component to chat room component. and also we no need to routes for team and one to one. because in api call for the get messages only **team_member_id** is the one not sending to team message, **team_id** is always sending. So if we pass selected chat object we can check chat type from it and modify api call in chat room component.

package.json
```diff
@@ -24,7 +24,9 @@
     "@ionic-native/core": "5.0.0-beta.21",
     "@ionic-native/splash-screen": "5.0.0-beta.21",
     "@ionic-native/status-bar": "5.0.0-beta.21",
-    "@ionic/angular": "4.0.0-beta.15",
+    "@ionic/angular": "4.0.0-beta.13",
+    "@ionic/storage": "^2.2.0",
```
shawnm0705: Don't need to use storage to save data to local storage now. Chaw is working on the storage service, we should implement the storage logic after chaw finished that service. That means we will use a shared service to save data into local storage, instead of importing "Storage" in every component

src/app/chat/chat.component.ts
```diff
@@ -1,8 +1,117 @@
...
+    }
+  }
+
+  getChatAvatarText(chatName) {
+    return this.chatService.generateChatAvatarText(chatName);
+  }
+
+  navigateToChatRoom(chat) {
+    this.storage.set('selectedChatObject', chat);
+    this.navCtrl.navigateForward('/chat/chat-room');
```
shawnm0705: Check the way we do navigation in other components, we need to make it consistent.

src/app/chat/chat.component.ts
```diff
@@ -1,8 +1,117 @@
...
+      this.haveMoreTeam = false;
+    }
+  }
+
+  getChatAvatarText(chatName) {
+    return this.chatService.generateChatAvatarText(chatName);
+  }
+
+  navigateToChatRoom(chat) {
+    this.storage.set('selectedChatObject', chat);
```
shawnm0705: It seems you just saved the chat object to local storage, and use this as a reference for the chat-room page.
The correct way of doing this is passing data in the route instead of save it to local storage. 
The route of the chat-room page should contain enough information to load the correct chat message. 
Only by doing that we can implement deep linking to every page (including the chat-room page)
Let's say we make the route of the chat-room like this `/chat/chat-room/:teamId/:teamMemberId`, because `team id` is necessary for every chat message. The logic is: if there's a `teamMemberId` passed in, it is a person to person message. If there's no `teamMemberId` passed in, it is a team message.

src/app/chat/chat.component.ts
sasangachathumal: i will change navigation, will implement the way that use in other components.

sasangachathumal: i change routing.

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -1,15 +1,142 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, ViewChild } from "@angular/core";
+import { Router, ActivatedRoute } from "@angular/router";
+import { Content } from "@ionic/angular";
+import { Storage } from "@ionic/storage";
```
shawnm0705: As mentioned above, use storage.service instead of this

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -1,15 +1,142 @@
...
+
+  goBack() {
+    this.router.navigateByUrl("/app/(chat:chat)");
+  }
+
+  sendMessage() {
+    if (this.message) {
+      const message = this.message;
+      this.message = "";
+      let data = {
```
shawnm0705: As we've discussed, don't hardcode in component anymore, build the full logic of component and hardcode data in service only

PR #25, opened 2018-11-13
Feature/av2 12/registration component

sasangachathumal:


src/app/auth/auth-registration/auth-registration.component.html
```diff
@@ -0,0 +1,35 @@
...
+        <ion-input name="password" [(ngModel)]="rePassword" required type="password" placeholder="Conform Password"></ion-input>
+      </ion-item>
+      <ion-item lines="none">
+        <ion-checkbox color="primary" slot="start" [(ngModel)]="isAgreed"></ion-checkbox>
+        <ion-label class="terms-and-conditions">I agree to the Terms & Conditions</ion-label>
+      </ion-item>
+    </ion-list>
+    <p><a (click)="openLink()">Terms & Conditions</a></p>
+    <p>You need a password! You need to agree with terms and Conditions.</p>
+    <ion-button (click)="login()" shape="round" expand="block" color="primary">REGISTER</ion-button>
```
shawnm0705: Where is the `login()` in component? Also, since it is 'register', call the function `register()` makes more sense

src/app/auth/auth-registration/auth-registration.component.html
```diff
@@ -0,0 +1,35 @@
...
+        <h4>Your Login Email Address</h4>
+        <p>{{user.email}}</p>
+    </ng-container>
+    <ng-container *ngIf="hide_password">
+        <h4>Your Mobile Number</h4>
+        <p>{{user.contact}}</p>
+    </ng-container>
+    <ion-list margin>
+      <ion-item>
+        <ion-input name="passowrd" [(ngModel)]="password" required type="password" placeholder="password"></ion-input>
```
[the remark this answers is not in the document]
sasangachathumal: placeholder is 'password'. you mean no need to use 'Conform Password', is that right.

src/app/auth/auth-registration/auth-registration.component.scss
```diff
@@ -0,0 +1,11 @@
+ion-input {
+    font-size: small !important;
```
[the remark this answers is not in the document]
sasangachathumal: screen shot 2018-11-16 at 2 31 37 pm
This is how UI looks like without changing font size. that's why change font size.
if make long text to two lines it not look good from user view. actually no need to put one word in another line. I don't know i'm toking about looking from user side. if we no need to make font size small i will change it to default. i try to make UI looks good to users.

src/app/auth/auth-registration/auth-registration.component.html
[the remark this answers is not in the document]
shawnm0705: I mean use capital letter first `P` instead of `p`

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  constructor() { }
+
+  ngOnInit() {
+  }
+
+  openLink() {
+    window.open('https://images.practera.com/terms_and_conditions/practera_default_terms_conditions_july2018.pdf', '_system');
+  }
+
```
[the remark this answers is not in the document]
shawnm0705: What I mean is to add a function to check the password match, that's why I didn't comment inside `openLink()`. I didn't mean this function

src/app/auth/auth.service.ts
```diff
@@ -21,6 +28,15 @@ export class AuthService {
...
-    });
+    }).pipe(map(response => {
+      console.log('Auth Response::', response);
+      this.storage.set('apikey', response.data.apikey);
+      return response;
+    }));
   }
+
+  saveRegistration(data): Observable<any> {
+    return of(true);
```
shawnm0705: use this as the mock API return, and component should respond based on "success" 
```
{
  "success": true
}
```

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,101 @@
...
+@Component({
+  selector: 'app-auth-registration',
+  templateUrl: './auth-registration.component.html',
+  styleUrls: ['./auth-registration.component.scss']
+})
+export class AuthRegistrationComponent implements OnInit {
+
+  password:String;
+  missingParams:boolean = true;
+  formValidationErrors:any = {
```
shawnm0705: Check https://angular.io/guide/reactive-forms#simple-form-validation and use the built in validator that Angular provided to do the form validation

src/app/auth/auth-registration/auth-registration.component.ts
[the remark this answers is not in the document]
shawnm0705: What I mean is to add a function to check the password match, that's why I didn't comment inside `openLink()`. I didn't mean this function

src/app/auth/auth-registration/auth-registration.component.html
```diff
@@ -0,0 +1,37 @@
...
+    <form (submit)="register()">
+      <ion-list margin>
+        <ion-item>
+          <ion-input name="passowrd" [(ngModel)]="password" (focus)="removeErrorMessages()" [ngModelOptions]="{standalone: true}" required type="password" placeholder="password"></ion-input>
+        </ion-item>
+        <ion-item>
+          <ion-input name="rePassword" [(ngModel)]="rePassword" (focus)="removeErrorMessages()" [ngModelOptions]="{standalone: true}" required type="password" placeholder="Conform Password"></ion-input>
+        </ion-item>
+        <ion-item lines="none">
+          <ion-checkbox color="primary" slot="start" (click)="validateRegistration('agree')" [(ngModel)]="isAgreed" [ngModelOptions]="{standalone: true}"></ion-checkbox>
```
shawnm0705: Don't need to do validation here, as mentioned above, use the Angular built in validator

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,101 @@
...
+    lengthError: false,
+    passwordMismatch: false
+  };
+  rePassword:String;
+  isAgreed:Boolean = false;
+  user:any = {
+    email: null,
+    activation_code: null
+  }
+
```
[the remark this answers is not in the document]
sasangachathumal: i implement that.

src/app/auth/auth.service.ts
```diff
@@ -192,4 +224,14 @@ export class AuthService {
...
\ No newline at end of file
+
+  saveRegistration(data: registerData): Observable<any> {
+    data.user_id = this.storage.get("hash").id;
+    data.key = this.storage.get("hash").key;
+    return this.request.post(api.register, data);
+  }
+
+  verifyRegistration(data: verifyParams): Observable<any> {
+    return this.request.post(api.verifyRegistration, data);
```
shawnm0705: This API call is not triggered, get error `ERROR TypeError: Cannot read property 'headers' of undefined` instead

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+    private utils: UtilsService,
+    private storage: BrowserStorageService,
+    private formBuilder: FormBuilder
+  ) {
+    this.initForm();
+  }
+
+  ngOnInit() {
+    this.domain =
+      this.domain.indexOf("127.0.0.1") !== -1
```
shawnm0705: add `localhost` as well

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+                break;
+              default:
+              this.errors.push(this.registerationForm.controls.errors[key]);
+            }
+            return;
+          }
+        }
+      }
+    }
+    // if (check) {
```
shawnm0705: What are the codes below? If we need to keep them, put some comment about the reason, otherwise remove them

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+  openLink() {
+    window.open(
+      "https://images.practera.com/terms_and_conditions/practera_default_terms_conditions_july2018.pdf",
+      "_system"
+    );
+  }
+
+  register() {
+    this.validateRegistration();
+    if ((this.registerationForm.valid) && (this.errors.length === 0)) {
```
shawnm0705: Since you already have a function to do the validation, `this.validateRegistration()` should return a boolean and be used here as the condition.

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+  selector: "app-auth-registration",
+  templateUrl: "./auth-registration.component.html",
+  styleUrls: ["./auth-registration.component.scss"]
+})
+export class AuthRegistrationComponent implements OnInit {
+  password: string;
+  confirmPassword: string;
+  isAgreed: boolean = false;
+  registerationForm: FormGroup;
+  hide_password: boolean = false;
```
shawnm0705: If I change it to true, it still ask me to fill the required field
image

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+      } else if (!this.isAgreed) {
+        this.errors.push("You need to agree with terms and Conditions.");
+      }
+    } else {
+      for (let conrtoller in this.registerationForm.controls) {
+        for (let key in this.registerationForm.controls[conrtoller].errors) {
+          if (this.registerationForm.controls[conrtoller].errors.hasOwnProperty(key)) {
+            switch (key) {
+              case "required":
+                this.errors.push("Fill required field");
```
shawnm0705: Since we don't have `*` to indicate the required field, it's better to change the message to "Please fill in your password"

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  openLink() {
+    window.open(
+      "https://images.practera.com/terms_and_conditions/practera_default_terms_conditions_july2018.pdf",
+      "_system"
+    );
+  }
+
+  register() {
```
shawnm0705: After I click "Register", nothing happened, console shows error instead

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,259 @@
...
+          .verifyRegistration({
+            email: this.storage.get("hash").email,
+            key: this.storage.get("hash").key
+          })
+          .subscribe(
+            response => {
+              let data = response.data.data;
+              let user = data.user;
+              let hash = this.storage.get("hash");
+              this.user.email = email;
```
shawnm0705: What is this?

src/app/auth/auth-registration/auth-registration.component.ts
sasangachathumal: i changed validation for this.

src/app/auth/auth-registration/auth-registration.component.ts
sasangachathumal: added * to placeholder and also message changed. because message meaningful in this way `Please fill in your password`

src/app/auth/auth-registration/auth-registration.component.ts
sasangachathumal: i added comments in the code to describe what is this.
i'm setting email and contact number after check registration valid and update storage data using new data came from api.

src/app/auth/auth.service.ts
sasangachathumal: The error i'm getting `Unexpected token u in JSON at position 0`. this error also getting in the login. I try to fixed it. but for that i need to edit requestService. so i think it's better chaw edit that. 
If you found any reason in my code please let me know.

src/app/auth/auth-registration/auth-registration.component.ts
sasangachathumal: same reason I mention before.

The error i'm getting `Unexpected token u in JSON at position 0`. this error also getting in the login. I try to fixed it. but for that i need to edit requestService. so i think it's better chaw edit that. 
If you found any reason in my code please let me know.

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -0,0 +1,278 @@
...
+          email: email,
+          key: activation_code
+        });
+        // check is Url valid or not.
+        this.authService
+          .verifyRegistration({
+            email: this.storage.get("hash").email,
+            key: this.storage.get("hash").key
+          }).subscribe(response => {
+              if (response) {
```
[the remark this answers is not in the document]
sasangachathumal: reason for this in request service error handler return the value as it’s a success value. So component error call backs are not calling because of that. I had a chat with caw and i have create a story for this issue. [`AV2-122`](https://intersective.atlassian.net/browse/AV2-122)

PR #34, opened 2018-11-27
[AV2-79]refactor activity page routing

shawnm0705:


trtshen:
looks good! @shawnm0705

PR #42, opened 2018-12-05
[AV2-111] feature/SharedModule

trtshen:


trtshen:
@shawnm0705 thanks for your help to find out the case sensitive `@Ionic/angular` issue. Both `@ionic/angular` and  `@Ionic/angular` are available in the ionic lib, we "just" need to pick the right one. With that fixed, i am now finally able to set it as "ready for review".

src/app/home/home.module.ts
```diff
@@ -1,21 +1,17 @@
...
+import { HomeRoutingModule } from './home-routing.module';
 import { NgCircleProgressModule } from 'ng-circle-progress';
-import { TodoCardComponent } from '../components/todo-card/todo-card.component';
-import { ActivityCardModule } from '../components/activity-card/activity-card.module';
-
 
 @NgModule({
   imports: [
-    IonicModule,
-    CommonModule,
```
shawnm0705: Since we've removed them here, we don't need the `import {} ` statement at the top as well

src/app/project/project.module.ts
```diff
@@ -8,16 +8,13 @@ import { NgCircleProgressModule } from 'ng-circle-progress';
...
 import { ProjectRoutingComponent } from './project-routing.component';
 import { ProjectRoutingModule } from './project-routing.module';
-import { ActivityCardModule } from '../components/activity-card/activity-card.module';
-
+import { SharedModule } from '@shared/shared.module';
 
 @NgModule({
   imports: [
-    IonicModule,
-    CommonModule,
```
shawnm0705: Same as above

src/app/app-routing.module.ts
```diff
@@ -18,17 +16,22 @@ const routes: Routes = [
     path: 'assessment', 
     loadChildren: './assessment/assessment.module#AssessmentModule',
   },
+  {
+    path: 'chat', 
+    loadChildren: './chat/chat.module#ChatModule' 
+  },
+  {
+    path: '',
```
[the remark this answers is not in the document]
trtshen: no it won't, you can try to run the app with my PR branch

trtshen:
thanks @shawnm0705 for the reviews!

PR #43, opened 2018-12-07
Feature/0.1/av2 56/reviews ui

ronakamani:


src/app/components/activity-card/activity-card.component.ts
```diff
@@ -1,20 +1,18 @@
...
 })
 export class ActivityCardComponent implements OnInit {
-  @Input() activity: object = {
-  	name: '',
-  };
-  name: string;
   
   constructor() {}
 
+  @Input() activity: {};
```
shawnm0705: Declare the type of these variables

src/app/reviews/reviews-routing.components.ts
```diff
@@ -0,0 +1,6 @@
+import { Component } from '@angular/core';
+
+@Component({
+	template: '<ion-router-outlet></ion-router-outlet>'
```
shawnm0705: Change tab to double spaces

src/app/reviews/reviews-routing.module.ts
```diff
@@ -0,0 +1,22 @@
+import { NgModule }             from '@angular/core';
+import { RouterModule, Routes } from '@angular/router';
+import { ReviewsComponent }    from './reviews.component';
+import { ReviewsRoutingComponent } from './reviews-routing.components';
+
+const reviewsRoutes: Routes = [
+  { 
+  	path: '',
```
shawnm0705: Change tab to double spaces

src/app/reviews/reviews.component.html
```diff
@@ -0,0 +1,23 @@
+<ion-header>
+  <ion-toolbar>
+    <ion-title center>Reviews</ion-title>
+  </ion-toolbar>
+</ion-header>
+
+<ion-content padding color="light">
+  <div class="top-round-btns" text-center>
+    <ion-button class="do" [ngClass]="{'active-review': activeToggle}" (click)="toggleActiveDo()">Todo</ion-button>
+    <ion-button class="done" [ngClass]="{'active-review': !activeToggle}" (click)="toggleActiveDone()">Done</ion-button>
```
shawnm0705: After click, we should disable this button. 
The problem of no doing that is currently if you click "Todo", and click "Todo" again, the button style changed, but the content doesn't

shawnm0705: Also, just a suggestion. 
"Toggle" usually means we can turn it on and off. Maybe consider change the name to "show" or "display" since the only behaviour we allow for clicking this button is to "display/show" todo/done page. 
Or if it is only one button, calling it "toggle" make sense that you can switch between two status

src/app/reviews/reviews.component.html
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+  <div *ngIf="!reviews.length; else reviewsCard" class="no-review-show" padding text-center>
+    <div class="empty-status" >
+      <h2 text-center>You have no new submissions to review</h2>
+    </div>
+  </div>
+  <ng-template #reviewsCard>
+    <ng-container *ngFor="let review of reviews">
+       <app-activity-card [review]="review" [toggle]="activeToggleDo"></app-activity-card>
```
shawnm0705: General suggestion (not only for this line):
The logic you used here is to pass `review` and `toggle` to `activity-card` and check whether to show this card or not in ActivityCardComponent. 
Since ActivityCardComponent is a separate component, what this component do should be just getting the necessary information and display it. "Whether show this card or not" this logic belongs to the review page only, so it should be placed here, not as part of the activity card component.
ActivityCardComponent should only handle the layout.

src/app/components/activity-card/activity-card.component.html
```diff
@@ -1,6 +1,49 @@
-<ion-card color="primary" class="card practera-card">
-  <ion-card-content padding>
-    <h4 color="primary-contrast">Activity</h4>
-    <h1 color="primary-contrast">{{name}}</h1>
+<ion-card 
+*ngIf="activity; else reviewCard"
```
shawnm0705: The reason of re-using a component is that we wanna use the same styling for different places. If we totally separate those layouts for `activity` and `review`, then there's no point that we use the same component.
That means we should 
1. Keep styling the same for main elements on the card.
2. Use a toggle to decide whether show this element or not.

shawnm0705: We can have a broader discussion with more people about which elements should be in the same styling. Please arrange that meeting if you think is necessary

src/app/home/home.component.ts
```diff
@@ -30,5 +30,9 @@ export class HomeComponent implements OnInit {
   goToActivity(id) {
     this.router.navigateByUrl('app/(project:activity/' + id + ')');
   }
+
+  goToReview(id) {
+    this.router.navigateByUrl('app/(project:activity/' + id +')');
```
shawnm0705: This goes to the wrong page

src/app/reviews/reviews.component.ts
```diff
@@ -0,0 +1,42 @@
...
+@Component({
+  selector: 'app-reviews',
+  templateUrl: './reviews.component.html',
+  styleUrls: ['./reviews.component.scss']
+})
+export class ReviewsComponent implements OnInit {
+
+  public reviews:Array <Review>;
+  public showDo: boolean = false;
+  public showDone: boolean = true;
```
shawnm0705: From our logic, display "Do" page and "Done" page can't happen at the same time, so probably we just need one boolean variable for this. Don't need `showDo` and `showDone` both

src/app/reviews/reviews.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    
+  };
+
+  activeDone (){
+    this.showDone = true;
+    this.showDo = false;
+    
+  };
+  gotoReview(id) {
+    this.router.navigateByUrl('app/(project:activity/' + id +')');
```
shawnm0705: That is the wrong page to go

src/app/reviews/reviews.component.scss
```diff
@@ -0,0 +1,25 @@
+
+.empty-status {
+  background: var(--ion-color-primary-contrast);
```
shawnm0705: there's a class in `style.scss` called `background-white`. Remove the CSS here and use this class instead

src/app/reviews/reviews.component.scss
```diff
@@ -0,0 +1,25 @@
...
+  .do {
+    --border-radius: 50px 0 0 50px;
+  }
+  .done {
+    --border-radius: 0 50px 50px 0;
+    margin-left: -4px;
+    
+  }  
+  .active-review {
+    --color: var(--ion-color-primary);;
```
shawnm0705: Use the directive `color="primary"` for this, do not create additional CSS

src/app/reviews/reviews.component.html
```diff
@@ -0,0 +1,53 @@
...
+  <ng-template #reviewsCard>
+    <ng-container *ngFor="let review of reviews">
+      <ion-card
+        color="primary"
+        class="card practera-card"
+        *ngIf="!review.isDone && !showDo" 
+        (click)="gotoReview(review.id)"
+      >
+        <ion-card-content class="review">
+          <h4 color="primary-contrast" class="activity-name">{{review.name}}</h4>
```
shawnm0705: `color=""` these kind of directives only works for ionic component, they don't work for normal html tags.
Also, the color of element that has a background color will use the `*-contrast` color. That means you don't need to specify it. Same for elements below

src/app/reviews/reviews.component.scss
ronakamani: we want to apply this color in a conditional situation, by getting this class. so it's not a permanent style, have to put in this special class.

src/app/home/home.component.html
```diff
@@ -26,7 +26,7 @@ <h1 text-center>{{ programName }}</h1>
 
     <ng-container *ngIf="todoItems.length; else noItems">
       <div *ngFor="let todoItem of todoItems">
-        <app-todo-card [todoItem]="todoItem"></app-todo-card>
+        <app-todo-card [todoItem]="todoItem" (click)="goToReview(todoItem.contextId,todoItem.id)"></app-todo-card>
```
shawnm0705: Check what is the "id" in the review routing, it's not the todo item id

shawnm0705: You put `todoItem.contextId` here, but there's no `contextId` in the home service. 
Also, since the `context_id` will be in the meta of todo item, it should be `todoItem.meta.contextId`

shawnm0705: But since these are related to home.service, we will change them when we implement home service. Will just leave it like this for now.

src/app/reviews/reviews.component.html
```diff
@@ -0,0 +1,53 @@
...
+      <h2 text-center>You have no new submissions to review</h2>
+    </div>
+  </div>
+  <ng-template #reviewsCard>
+    <ng-container *ngFor="let review of reviews">
+      <ion-card
+        color="primary"
+        class="card practera-card"
+        *ngIf="!review.isDone && !showDo" 
+        (click)="gotoReview(review.contextId,review.id)"
```
shawnm0705: Check what is the "id" in review routing, it's not the review.id

src/app/reviews/reviews.component.html
```diff
@@ -0,0 +1,53 @@
...
+          </h2>
+          <ion-text >{{review.submittedReviewDate}}</ion-text>
+          
+        </ion-card-content>  
+      </ion-card>
+      <ion-card
+        color="primary"
+        class="card practera-card"
+        *ngIf="review.isDone && showDo"
+        (click)="gotoReview(review.contextId,review.id)"
```
shawnm0705: Same as above here

shawnm0705:
@ronakamani I'll merge this PR now, since home service need some of these changes. The problem I mentioned in the last review will occur when we implement the reviews service. We can solve them at that time. Or if you wanna solve them now, please create a new PR.

PR #45, opened 2018-12-12
chat-routing

trtshen:


trtshen:
hi @sasangachathumal @shawnm0705 I would recommend us to handle the chat routing in this way.

PR #51, opened 2018-12-14
[AV2-87] Setting service

trtshen:
- add contanct number to user storage data
- setting service instatiation


src/app/switcher/switcher.service.ts
```diff
@@ -15,9 +15,10 @@ const api = {
 };
 
 export interface ProgramObj {
-  program: Program;
-  project: Project;
-  timeline: Timeline;
+  program: Program,
```
trtshen: @zawmyohtut for interface, the comma should be semicolon ";"  it's not object.

zawmyohtut: @trtshen hi chaw, I have updated. and will merge the PR now.

PR #53, opened 2018-12-17
Feature/0.1/av2 85/topic-service

ronakamani:


src/app/topic/topic.service.ts
```diff
@@ -1,66 +1,119 @@
...
+  
+  getTopicProgress(topicId): Observable<any> {
+    return this.request.get(api.get.progress, {params: {
+        model: 'Activity',
+        model_id: topicId,
+        scope: 'Task'
+      }})
+      .pipe(map(response => {
+        if (response.success && response.data) {
+          var progress = response.data.Activity.Topic.find(function (topic) {
```
shawnm0705: return data format validation missing

src/app/topic/topic.service.ts
```diff
@@ -1,66 +1,122 @@
...
+        scope: 'Task'
+      }})
+      .pipe(map(response => {
+        if (response.success && !this.utils.isEmpty(response.data)) {
+
+            var progress = response.data.Activity.Topic.find(function (topic) {
+                return topic.id === topicId;
+            }) 
+            this.topicProgress = progress.progress;
+            return this.topicProgress;
```
shawnm0705: If we only use `this.topicProgress` for the return the data, why do we need this variable?

src/app/topic/topic.service.ts
```diff
@@ -1,66 +1,122 @@
...
-        
-        },
-      ]
-    };
-  
-  topicDone = {
-    1: false,
-    2: false
-  }
+  topic :Topic;
```
shawnm0705: Did we use this variable anywhere?

PR #54, opened 2018-12-17
[AV2-91]home service

shawnm0705:


src/app/home/home.component.ts
```diff
@@ -1,27 +1,54 @@
...
-    private homeService: HomeService 
+    private homeService: HomeService,
+    private utils: UtilsService
   ) {}
 
   ngOnInit() {
-    this.todoItems = this.homeService.getTodoItems();
-    this.activity = this.homeService.getCurrentActivity();
+    this.homeService.getTodoItems()
+      .subscribe(todoItems => {
```
trtshen: @shawnm0705 getTodoItems() in develop branch is no longer return observable, could you please check?
https://github.com/intersective/practera-app-v2/blob/develop/src/app/home/home.service.ts#L49-L55

PR #59, opened 2018-12-20
Feature/0.1/av2 83/chat services

sasangachathumal:


src/app/chat/chat.service.ts
```diff
@@ -128,29 +160,39 @@ export class ChatService {
...
-      sender_name: 'Chathumal',
-      is_sender: true,
-      team_name: 'dream team',
-      sent_time: '1.30 PM',
-      message: data.message,
-      team_id: data.team_id,
-      to: data.to
-    };
-    return of(returnData);
+    const body = new HttpParams()
```
shawnm0705: Don't need to create a HttpParams. We can just pass an object of body parameters to `this.request.post()`

src/app/chat/chat.service.ts
```diff
@@ -128,29 +160,39 @@ export class ChatService {
...
   markMessagesAsSeen(prams: markAsSeenPrams): Observable<any> {
-    prams.action = 'mark_seen';
-    return of("maked");
+    let body = {
+      'team_id': prams.team_id,
+      'id': prams.id,
+      'action': 'mark_seen'
+    }
+    return this.request.post(api.markAsSeen, body, {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
```
shawnm0705: Can we use the default content-type `application/json` for the header of this API call? If so, don't need to specify it here. Same for other places

src/app/chat/chat-list/chat-list.component.ts
```diff
@@ -36,6 +36,7 @@ export class ChatListComponent implements OnInit {
    * @param {Array} response 
    */
   private updateChatListResponse(response):void {
+    console.log('updateChatListResponse', response)
```
shawnm0705: Remove this if not needed

src/app/home/home.service.ts
```diff
@@ -154,13 +154,7 @@ export class HomeService {
...
     let unreadMessages = 0;
     let noOfChats = 0;
-    let todoItem: TodoItem = {
-      type: 'chat',
-      name: '',
-      description: '',
-      time: '',
-      meta: {}
-    };
+    let todoItem: TodoItem;
```
shawnm0705: Why do we change this here?

src/app/chat/chat.service.ts
```diff
@@ -128,29 +160,39 @@ export class ChatService {
...
-      message: data.message,
-      team_id: data.team_id,
-      to: data.to
-    };
-    return of(returnData);
+    const body = new HttpParams()
+      .set('data[to]', data.to.toString())
+      .set('data[message]', data.message)
+      .set('data[team_id]',data.team_id.toString())
+      .set('env', 'develop');
```
shawnm0705: Why do we pass the `env` as `develop`?

src/app/chat/chat.service.ts
```diff
@@ -30,8 +58,12 @@ export class ChatService {
...
-    this.initDemoData();
+  constructor(
+    private request: RequestService,
+    private storage: BrowserStorageService,
+    private utils: UtilsService
+  ) {
+    
   }
 
   initDemoData() {
```
shawnm0705: Remove the demo data if not need

src/app/chat/chat-list/chat-list.component.ts
```diff
@@ -36,6 +36,7 @@ export class ChatListComponent implements OnInit {
    * @param {Array} response 
    */
   private updateChatListResponse(response):void {
```
shawnm0705: This is a function for data normalisation, should be moved to service file

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -19,57 +19,60 @@ interface Chat {
...
-  ) {
+  ) {}
+
+  ngAfterContentInit() {
+    this.content.scrollToBottom();
   }
 
   ngOnInit() {
     this.chatColors = this.storage.get("chatAvatarColors");
     this.selectedChat = this.storage.get("selectedChatObject") || {
```
shawnm0705: Why do we have to get & set the chat object in local storage? 
Can we pass those as parameters in the route of this page?
Note:
Think about how we implement direct link to the chat room page.

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -19,57 +19,60 @@ interface Chat {
...
-      team_id : this.routeTeamId,
+      team_id: this.routeTeamId,
       page: this.getMessagePageNumber(),
       size: this.messagePagesize,
       team_member_id: this.routeTeamMemberId
-    }
+    };
     this.chatService.getMessageList(param).subscribe(response => {
-      this.updateMessageListResponse(response, false);
+      this.updateMessageListResponse(response.data, false);
```
shawnm0705: This function is the data normalisation, move it to the service file.

src/app/home/home.service.ts
sasangachathumal: because in dashboard always showing one empty chat notification card if there are no todoItem to show.

src/app/chat/chat.service.ts
sasangachathumal: we need to set env by checking the URL if it is sandbox we send develop, if it live we pass another thing. reason we passing that we back end need that env to identify the pusher event. in here i miss that URL checking part. i will fix that.

src/app/chat/chat.service.ts
sasangachathumal: for this api request we need to pass those data in body as Form data.. if we didn't pass it not working. 
i try to pass data as params and pass empty object to request body. but didn't work. this is also happening in login.

src/app/chat/chat.service.ts
sasangachathumal: yes,`application/json` can use as default content-type, but in hear this api calls not working if we didn't pass `application/x-www-form-urlencoded` as the content-type. I have try to remove it. in the way request service if we need and header pass with api we need to pass it to request service if we didn't pass any header default header will set to api request. this content-type is only need in some api request so when we calling those api request from service we need to pass this custom content-type to request service to set request header.

src/app/chat/chat-room/chat-room.component.ts
sasangachathumal: i try some options,
1.  passing data using `this.router.navigate`. method using now.
2. passing data using `this.navCtrl.navigateForward`.
in both option to pass chat object i have use `JSON.stringify()` method to before add chat object to param. and add `JSON.parse()` method to access data. but issue is after `JSON.stringify()` and set to parameters chat object data showing on the URL. 
BUT I have use that method because URL not showing the PWAs. but if some one use browser it will show the chat object data.

src/app/chat/chat-room/chat-room.component.html
```diff
@@ -31,10 +36,11 @@
...
-    <ion-item lines="none">
-        <ion-textarea placeholder="Enter your message" (keydown.enter)="sendMessage($event)" required="true" name="message" [(ngModel)]="message"></ion-textarea>
-      <ion-button (click)="sendMessage()" type="submit" slot="end" fill="clear" color="light">
-        <ion-icon name="send"></ion-icon>
-      </ion-button>
-    </ion-item>
-</ion-footer>
\ No newline at end of file
+  <ion-item lines="none">
+    <ion-textarea placeholder="Enter your message" (keydown.enter)="sendMessage($event)" required="true" name="message"
```
shawnm0705: Pressing enter should go to a new line instead of send the message

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -95,58 +162,27 @@ export class ChatRoomComponent implements OnInit {
...
-    let chatcolor = chatColors.find(function(chat) {
-      return chat.name === response[index].sender_name;
-    });
-    if (chatcolor) {
-      response[index].chat_color = chatcolor.chat_color;
-    } else {
-      response[index].chat_color = this.chatService.getRandomColor();
+      this.chatService.postNewMessage(data).subscribe(
+        response => {
+          this.messageList.push(response.data);
```
shawnm0705: User should be able to see the message he/she is sending directly without delay.
We can display a spinner besides the message indicating that it is sending to the server.

src/app/chat/chat.service.ts
```diff
@@ -128,29 +91,74 @@ export class ChatService {
...
+      .set("env", environment.env);
+      if (data.participants_only) {
+        body.set("participants_only", data.participants_only.toString());
+      }
+    return this.request.post(api.createMessage, body.toString(), {
+      headers: { "Content-Type": "application/x-www-form-urlencoded" }
+    });
+  }
+
+  unreadMessageCout(data: unreadMessagePrams): Observable<any> {
```
shawnm0705: We are not using this anywhere

src/app/chat/chat.service.ts
```diff
@@ -178,4 +186,203 @@ export class ChatService {
   private getRamdomNumber() {
```
shawnm0705: We need a way to make sure two users doesn't use the same color. This should be inside `getRandomColor()`

src/app/chat/chat-room/chat-room.component.html
sasangachathumal: This is the way we implement this logic in app v1 and also this is the way every other chat apps  web version using. Why we need to change this?

src/app/chat/chat.service.ts
sasangachathumal: I have create this method it use to show unread count in tab in app v1 and also use to update unread count with pusher in app v1. that's why i create this function if it is no need to use after pusher implemented i will remove it.

src/app/chat/chat.service.ts
sasangachathumal: The logic is selecting random color two users can have same color to identify those two that's why we using Avatar text. This logic similar to android contact app. This is the way we use in app v1 and this is the way designers also agreed. we have 19 colors that designers selected we are using one random color from them.

src/app/chat/chat.service.ts
```diff
@@ -178,4 +186,203 @@ export class ChatService {
...
+  /**
+   * this method check old avatar colors and set the releted one.
+   * if there no old avatar colors it will create new color and add that to service variable.
+   * @param {Array} response
+   * @param {Array} chatColors
+   * @param {String} status
+   */
+  private setChatAvatarColors(response, chatColors, status): Array<any> {
+    let index = 0;
+    for (index = 0; index < response.length; index++) {
```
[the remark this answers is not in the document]
sasangachathumal: Array.forEach() run in synchronously so because of that to execute some function or code after for each we have track it's index using variables or we have use promises to check is loop is over. and also Array.forEach is bit more costly when compair with for loops. in here i use for loop it's easy to get to know loop is over and execute function after loop over.

src/app/chat/chat-list/chat-list.component.html
```diff
@@ -14,7 +19,7 @@
         </ion-avatar>
         <ion-label>
           <div class="chat-name-container">
-            <h2>{{chat.is_team ? chat.team_name : chat.name}}</h2>
+            <h2 [class]="chat.is_team ? 'chat-name team-channel' : 'chat-name'">{{getChatName(chat)}}</h2>
             <ng-container *ngIf="(chat.role === 'mentor') && (haveMoreTeam)"><span>mentor - {{chat.team_name}}</span></ng-container>
             <ng-container *ngIf="(chat.role === 'mentor') && (!haveMoreTeam)"><span>mentor</span></ng-container>
             <ng-container *ngIf="(chat.role !== 'mentor') && (!chat.is_team) && (haveMoreTeam)"><span>{{chat.team_name}}</span></ng-container>
```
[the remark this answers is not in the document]
sasangachathumal: no. this mean chat role is not mentor and it's not a team chat (that's mean one to one chat) if we have more than one team showing the team name after chat name.

src/app/chat/chat.service.ts
```diff
@@ -128,29 +91,74 @@ export class ChatService {
...
+          }
+        })
+      );
+  }
+
+  private _normalisTeamResponse(response) {
+    return response.Team.name;
   }
 
   generateChatAvatarText(text) {
```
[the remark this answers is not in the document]
sasangachathumal: As avatar text will show first two letters of `chatNameArray[1]`.
if `chatNameArray[0]` have value and `chatNameArray[1]` null. it will get first two letters of `chatNameArray[0]`.

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -12,77 +12,144 @@ interface Chat {
...
+            tempRes.reverse();
+            if (loadMore) {
+              this.messageList = tempRes.concat(this.messageList);
+            } else {
+              this.messageList = tempRes;
+            }
+            this.getChatName();
+            this.markAsSeen(tempRes);
+          } else {
+            this.messagePageNumber -= 1;
```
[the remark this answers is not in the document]
sasangachathumal: i do this to stop increasing page number if there are no response. When we call message list api page number equals to 2 but there are no message in response. we need to reduce the page number. if we didn't do that in next api page number will be 3.

src/app/chat/chat.service.ts
```diff
@@ -178,4 +186,203 @@ export class ChatService {
...
+        }
+    }
+    }
+    return this.chatList;
+  }
+
+  private _normalisMessageListResponse(response, selectedChat): Array<any> {
+    let index = 0;
+    if (response.length > 0) {
+      for (index = 0; index < response.length; index++) {
```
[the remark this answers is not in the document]
sasangachathumal: why i use forEach i have mention before.

src/app/shared/request/request.interceptor.ts
```diff
@@ -34,7 +34,10 @@ export class RequestInterceptor implements HttpInterceptor {
     }
 
     // do not need to pass team id for teams.json
-    if (teamId && !req.url.includes('/teams.json')) {
+    // do not need to pass team id for chat api calls
+    if (teamId && !req.url.includes('/teams.json') && 
+    !req.url.includes('/message/chat/list.json') && !req.url.includes('/message/chat/create_message') &&
```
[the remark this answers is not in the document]
sasangachathumal: this can be done when user in one team but there is a way that user in multiple teams in that case i need to pass team ID to api in body. that's why i pass team ID in body.

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -12,77 +12,144 @@ interface Chat {
...
+        if (response) {
+          if (response.length > 0) {
+            tempRes = Object.assign([], response);
+            tempRes.reverse();
+            if (loadMore) {
+              this.messageList = tempRes.concat(this.messageList);
+            } else {
+              this.messageList = tempRes;
+            }
+            this.getChatName();
```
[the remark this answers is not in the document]
sasangachathumal: can't do it in normalization function because i need to set chat name in to selected chat object not to message object. message object have a name.

src/app/chat/chat.service.ts
```diff
@@ -178,4 +186,203 @@ export class ChatService {
...
+          this.chatList.splice(index,1);
+        }
+    }
+    }
+    return this.chatList;
+  }
+
+  private _normalisMessageListResponse(response, selectedChat): Array<any> {
+    let index = 0;
+    if (response.length > 0) {
```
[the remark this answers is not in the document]
sasangachathumal: i have check API return data before call normalise function.

src/app/chat/chat-room/chat-room.component.html
shawnm0705: The reason is that on mobile you don't have "Shift+Enter"/"Command+Enter"

src/app/chat/chat.service.ts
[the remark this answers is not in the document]
shawnm0705: Firstly, Array.forEach() is not asynchronous, so don't need to use promise for that. Try this:
```
let b = [1,2,3,4,5,6];
b.forEach((v, i) => {
  b[i]++;
});
console.log(b);
```

Secondly, regarding the performance of `for` and `forEach()`, `for` is faster than `forEach()`, but the difference is too little that we can just omit. There're benefits of using `forEach()` as well, such as readability.

So, the reason I pointed this out is that I think it's better to make it consistent among the codebase. Since we use `forEach()` in all other places, it's better to use it here as well.

Anyway, there's no much difference regarding the functionality.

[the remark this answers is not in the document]
sasangachathumal: yes sure. I will change for to foreach.

src/app/chat/chat-room/chat-room.component.html
sasangachathumal: we need to ask this from anu and beenish before change.

PR #60, opened 2018-12-20
[AV2-99]refactor tabs component

shawnm0705:


src/app/tabs/tabs.component.ts
```diff
@@ -1,11 +1,34 @@
...
+  showChat = true;
+  noOfTodoItems = 0;
+  noOfChats = 0;
+
+  constructor (
+    private tabsService: TabsService,
+    private storage: BrowserStorageService
+  ) {}
+
+  ngOnInit() {
```
trtshen: @shawnm0705 ngOnInit() will only initiated once during class instantiation. And i saw the `noOfTodoItems` code placed inside ngOnInit, so if updating the number of item/chats once per app load is intended? If they aren't you can consider using ionic's `ionViewWillEnter` lifecycle hook

PR #64, opened 2019-01-03
[AV2-120]login fail redirect to program selection

shawnm0705:


trtshen:
I'll spend some more time to find out what is blocking pressing "enter" button to submit form. by default that's the default behavior of html form, adding the `keyup.enter` watcher has disabled most of the form validation.

trtshen:
As you've mention, it's ionic4 issue. Let's get this merged first, and go back to the default "enter" to submit behavior for later.

PR #86, opened 2019-01-07
circular dependent fix

trtshen:
Minor code changes
- NotificationService.popUp's second parameter can be empty and set false as default
- reviewRating popup is better place at assessment.service, because only assessment.component will use it. (it is more reuseable if we keep them serve the generic purpose)


src/app/review-rating/review-rating.service.ts
```diff
@@ -22,6 +23,7 @@ export class ReviewRatingService {
 
   constructor(
   	private request : RequestService,
+    private notification : NotificationService,
```
shawnm0705: We didn't use it here

src/app/assessment/assessment.component.ts
```diff
@@ -72,7 +73,8 @@ export class AssessmentComponent implements OnInit {
     private assessmentService: AssessmentService,
     private utils: UtilsService,
     private notificationService: NotificationService,
-    private storage: BrowserStorageService    
+    private storage: BrowserStorageService,
+    private reviewRatingService: ReviewRatingService,
```
shawnm0705: We didn't use it here

src/app/shared/notification/notification.service.ts
```diff
@@ -15,18 +14,33 @@ export class NotificationService {
...
-      }
-    });
+  async popUp(type, data, redirect:any = false) {
+    const component = PopUpComponent;
+    const componentProps = {
+      type,
+      data,
+      redirect,
+    };
+    const modal = await this.modalController.create(this.modalConfig({ component, componentProps }));
```
shawnm0705: We can use `this.modal()` here since it is defined below

trtshen:
thanks @shawnm0705 for reviewing!

PR #89, opened 2019-01-07
[AV2-130]implement file preview functionality of filestack

shawnm0705:


src/app/shared/filestack/filestack.service.ts
```diff
@@ -32,7 +36,27 @@ export class FilestackService {
...
 
   //get s3 config
-  getS3Config () {
+  getS3Config() {
     return this.s3Config;
   }
+
+  previewFile(file) {
+    let fileUrl = file.url;
+    if (fileUrl && fileUrl.indexOf('api/file') !== -1) {
```
trtshen: @shawnm0705 there are some updates in filestack. I've just checked the latest preview format and i think we should follow the latest one.
for example, `https://cdn.filestackcontent.com/preview/SGe380ViQ7W8efiX6sdG`

have a look at the docs too. https://www.filestack.com/docs/concepts/transform/#document-viewer

PR #97, opened 2019-01-08
Feature/0.1/av2 84/project service

ronakamani:


src/app/project/project.service.ts
```diff
@@ -1,142 +1,223 @@
...
+  }
+  
+  private _normaliseMilestones(data){
+    
+    if (!Array.isArray(data)) {
+      this.request.apiResponseFormatError('Milestones array format error');
+      return [];
+    }
+    this.milestones = [];
+    this.milestone_ids = [];
```
shawnm0705: Don't need to define `milestones` and `milestone_ids` as global variables since we just use those within this function.
Note:
Do not define global variables if we don't have to. This will cause the problem below:
1. Use more memory to store the global variable
2. The problem that you've face before. Because these are global variables, after switch to activity page, `milestones` still exist with the previous value. If you didn't assign empty array here, milestones will be added upon the previous one. If we didn't define these as global variable, those variable won't exist anymore when you come back from activity page, so that this problem won't happen
3. Also, it is not clear that where have we touched this global variable

src/app/project/project.service.ts
```diff
@@ -1,142 +1,223 @@
...
+      this.milestone_ids.push(milestone.id);
+    }
+  });
+
+  this._getActivities().subscribe(activities => { 
+    
+    this.activities = activities; 
+    this._addActivitiesToEachMilestone(this.milestones, this.activities);
+    this.milestone_ids.forEach(id => {
+      this._getProgress(id, this.milestones).subscribe();
```
shawnm0705: We don't need to get the progress within the loop of milestones. Otherwise we will call this API multiple times to get those progresses. Instead, we can just call it once to get all progresses. Also, the scope shouldn't be "Task" since we only need the progress of activities.

src/app/project/project.service.ts
```diff
@@ -1,142 +1,223 @@
...
-          name: 'Test Activity six',
-          isLocked: false,
-          progress: 1,
-          hasFeedback: false,
-        },
-      ],
-      isLocked: false
+     });
+  });
+}
```
shawnm0705: Indentation

src/app/project/project.service.ts
```diff
@@ -1,142 +1,223 @@
...
+      milestone.id = eachMilestone.id;
+      milestone.name = eachMilestone.name;
+      milestone.isLocked = eachMilestone.is_locked;
+      if (this.utils.has(eachMilestone, 'description')) {
+        milestone.description = eachMilestone.description;
+      };
+      
+      this.milestones.push(milestone);
+      
+  });
```
shawnm0705: Indentation

src/app/project/project.service.ts
```diff
@@ -1,142 +1,223 @@
...
+    });
+    
+  });
+  
+  return this.milestones;
+  }
+
+  private _addActivitiesToEachMilestone(milestones,activities) {
+    
+    milestones.forEach(function(eachMilestone) {
```
shawnm0705: Don't use this two layer loop. Check `_.find()` in lodash see if it helps

src/app/project/project.service.ts
```diff
@@ -1,142 +1,223 @@
...
 
 export interface Activity {
-  id?: number;
-  name?: string;
-  isLocked?: boolean;
-  progress?: number;
-  hasFeedback?: boolean;
+  id: number;
+  name: string;
+  milestoneId?: number;
```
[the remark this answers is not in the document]
ronakamani: we  assign Activity to each milestone based on milestonId. I used this variable in the code.

PR #115, opened 2019-01-09
[AV2-75] Multiple lines on chat

sasangachathumal:
create a pipe to identify urls.
added text area to type message.
added some styles to show multi line messages.
created some new css variables.


sasangachathumal:
This is have RUL identify pipe but it not use in this feature. because it have some issue with anchor tag color. it will be done in another story.

src/app/home/todo-card/todo-card.component.scss
```diff
@@ -5,6 +5,19 @@
   background: #fff;
   min-height: 58px;
 }
+.todo-item {
```
shawnm0705: This class has been removed already. Do we still need this CSS ?

src/theme/variables.scss
```diff
@@ -5,6 +5,9 @@
 
 :root {
 
+  --practera-black: black;
```
shawnm0705: We don't need to create a variable for this, just use `black` & `white` in CSS

src/app/home/todo-card/todo-card.component.scss
sasangachathumal: I use this to limit long description text in the todo items on dashboard.

sasangachathumal: it was there i didn't add this. I have added this part.
`.icon-item {
    ion-text {
      p {
        max-height: 16px;
        -webkit-line-clamp: 1;
        line-height: 16px;
        overflow: hidden;
      }
    }
  }`

PR #183, opened 2019-01-23
[AV2-187]/add timeformater to the review timestamp

ronakamani:


src/app/reviews/reviews.service.ts
```diff
@@ -63,12 +63,19 @@ export class ReviewsService {
...
         submitterName: review.AssessmentSubmission.Submitter.name,
-        date: review.AssessmentReview.is_done ? review.AssessmentReview.modified : review.AssessmentReview.created,
+        date: this._timeFormater(review.AssessmentReview.is_done ? review.AssessmentReview.modified : review.AssessmentReview.created),
         teamName: this.utils.has(review, 'AssessmentSubmission.Team.name') ? review.AssessmentSubmission.Team.name : '',
         contextId: review.AssessmentSubmission.context_id
       });
     });
     return reviews;
   }
+  private _timeFormater(time: string) {
```
shawnm0705: Please move this function to util service since we are using it in two places

shawnm0705:
@ronakamani Please check the message I mentioned above

> The problem of this bug is that the way we re-format the date does not work on Safari. Please fix it

PR #225, opened 2019-02-04
[AV2-210]/fix loading the current activity when progress is 1

ronakamani:


src/app/home/home.service.ts
```diff
@@ -245,7 +245,7 @@ export class HomeService {
       this.request.apiResponseFormatError('Progress.Milestone.Activity format error');
       return ;
     }
-    if (activity.progress < 1) {
+    if (activity.progress < 1 || activity.progress ===1) {
```
shawnm0705: The logic is incorrect, if we do so, then we will regard the first activity that has progress 1 as the current activity.
The correct logic should be "if there's no activity that has progress < 1, regard the last activity as the current activity"

PR #230, opened 2019-02-05
[AV2-194]/milestone scroll

ronakamani:


src/app/project/project.component.html
```diff
@@ -51,7 +51,7 @@
                 <ng-container *ngIf="loadingActivity" >
                   <ion-spinner></ion-spinner>
                 </ng-container>
-                <ion-item class="activity-item" *ngFor="let activity of milestone.Activity" color="light">
+                <ion-item class="activity-item" *ngFor="let activity of milestone.Activity" color="light" ng-class="addMargin: 'test'?">
```
sasangachathumal: To change class we using `[ngClass]`. I think we need to change this `ng-class` to `[ngClass]`. And there are no css call `addMargin`

src/app/project/project.component.html
```diff
@@ -78,4 +78,5 @@
       </div>
     </ion-item>
   </ion-list>
+  <div style="height:300px"></div>
```
sasangachathumal: We need move this styles to `project.component.scss` file.

src/app/project/project.component.ts
```diff
@@ -40,6 +40,7 @@ export class ProjectComponent extends RouterEnter {
   @ViewChildren('milestoneRef', {read: ElementRef}) milestoneRefs: QueryList<ElementRef>;
   private milestonePositions: Array<number> = [];
   public activeMilestone: Array<boolean> = [];
+  public test: boolean = false;
```
sasangachathumal: Need to use proper name instead of `test`.

PR #274, opened 2019-02-18
AV2-240/badges on dashboard

ronakamani:


src/app/home/home.component.ts
```diff
@@ -66,8 +68,14 @@ export class HomeComponent extends RouterEnter {
     this.loadingTodoItems = true;
     this.loadingProgress = true;
     this.loadingActivity = true;
+    this.achievements= [];
     // add a flag in local storage to indicate that is there any fast feedback open
     this.storage.set('fastFeedbackOpening', false);
+
+    this.achievementService.getAchievements('desc').subscribe(achievements => {
+      this.achievements = achievements;
+      this.achievements.length = 3;
```
shawnm0705: This is not enough. We need the logic to display certain 3 achievements. (e.g. 2 earned, 1 not earned, these logics)

src/styles.scss
```diff
@@ -132,3 +132,21 @@ body {
 .no-bg {
   background-color:  transparent !important;
 }
+
+.achievement {
```
shawnm0705: These CSS are not shared ones, we shouldn't put those in `styles.scss`

PR #295, opened 2019-02-26
[AV2-98]event list page

shawnm0705:


src/app/events/events.service.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  private _normaliseEvents(data): Array<Event> {
+    if (!Array.isArray(data)) {
+      this.request.apiResponseFormatError('Event format error');
+      return [];
+    }
+    let events: Array<Event> = [];
+    data.forEach(event => {
+      if (!this.utils.has(event, 'id') ||
```
trtshen: hi @shawnm0705 , just wondering if this checking for each field is really necessary? 
1) we could add the Event interface for the "event" param above, for eg.
```javascript
data.forEach(event: Event => {
// ...
});
```

2) I think it'll be better if we can just trust API is returning correct API response format, but it's open for discussion.

shawnm0705: @trtshen I think we can't just put the interface as the type, because API could return more than we need. Since we are doing the data normalisation, I think it's better to do normalisation for every API.
By checking each field, if we change API returns later, we will be able to find out if it still works for App or not. 
By always doing this kind of check and re-assigning data, we can make sure that even we add more return data for API later, it will still work. What do you think?

shawnm0705: Also if you have noticed, I've changed the variable name from `a_b` to `aB` camelCase, so that we always use camelCase variables

shawnm0705: Also, this actually gives a list of fields that we are using in the App.
Also if you've noticed that I've converted snake_cased variables to camelCase ones, so that we always use camelCase variables in the code. Sometimes we use different names as well

PR #297, opened 2019-02-26
[AV2-228]/fix the setting page issue

ronakamani:


src/app/settings/settings.component.ts
```diff
@@ -231,4 +231,13 @@ export class SettingsComponent extends RouterEnter {
     return this.authService.logout();
   }
 
+  setCursorPosition(event) {
+    event = (event) ? event : window.event;
+    var charCode = (event.which) ? event.which : event.keyCode;
+    if (charCode > 31 && (charCode < 48 || charCode > 57)) {
```
shawnm0705: Please make some comments on what are those numbers

src/app/settings/settings.component.ts
```diff
@@ -231,4 +231,13 @@ export class SettingsComponent extends RouterEnter {
     return this.authService.logout();
   }
 
+  setCursorPosition(event) {
```
shawnm0705: I didn't get what is this function doing? what does `return true;` / `return false;` mean?

PR #309, opened 2019-03-01
[AV2-261] Auto saving feature

sasangachathumal:
- update assessment service saveAnswer method with new parameters.
- update assessment component show saving message and buttons.
- pass save changes methods to question components as output.
- add some styles to subtitle and text button.


src/app/assessment/assessment.service.ts
```diff
@@ -408,20 +423,25 @@ export class AssessmentService {
     return answer;
   }
 
-  saveAnswers(assessment, answers, action) {
+  saveAnswers(params: SaveAnswersParams) {
```
shawnm0705: We don't need to create this parameters as an Interface, we can just put `submissionId` as optional parameter

shawnm0705: BTW, `AssessmentSubmission`, `AssessmentReview` these are all names that we used in Practera backend for naming convention. Here is App we can just use `assessment`, `submission`, `review` for short

src/app/assessment/assessment.component.ts
```diff
@@ -240,37 +277,59 @@ export class AssessmentComponent extends RouterEnter {
       });
     }
 
+    const params: SaveAnswersParams = {
+      assessment: assessment,
+      answers: answers,
+      action: this.action
+    };
+    if (this.submission) {
+      params.AssessmentSubmissionId = this.submission.id;
```
shawnm0705: As I mentioned in assessment.service, we don't need to create it as an interface. 
Here we can just initialise `submissionId` as `null`, if there is a submission, pass the actual submission id

src/app/assessment/assessment.component.ts
```diff
@@ -192,19 +209,33 @@ export class AssessmentComponent extends RouterEnter {
     return requiredQuestions;
   }
 
-  submit() {
-    this.submitting = true;
+
+
+  submit(isInProgress: boolean) {
```
shawnm0705: I think `saveInProgress` is better than `isInProgress`

src/app/assessment/assessment.component.ts
```diff
@@ -240,37 +277,59 @@ export class AssessmentComponent extends RouterEnter {
...
-            {
-              text: 'OK',
-              role: 'cancel',
-              handler: () => {
-                this.router.navigate(['app', 'home']);
-                return;
+        this.saving = false;
+        if (isInProgress) {
+          // display message for successfull saved answers
+          this.savingMessage = 'Last saved a mooment ago';
```
shawnm0705: `mooment`

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
shawnm0705: Is this file needed? Otherwise maybe put it in `.gitignore`

src/app/assessment/assessment.component.html
```diff
@@ -2,14 +2,16 @@
   <ion-toolbar color="light">
     <ion-icon name="arrow-back" (click)="back()" color="primary" slot="start" margin-start></ion-icon>
     <ion-title text-center>Assessment</ion-title>
+    <ion-title *ngIf="saving || doAssessment || doReview" text-center class="subTitle">{{savingMessage}}</ion-title>
```
shawnm0705: I think we don't need to put `saving` as the condition of this, it will always be displayed for `doAssessment` or `doReview`

src/app/assessment/assessment.component.html
```diff
@@ -2,14 +2,16 @@
...
+    <ion-button *ngIf="doAssessment || doReview" [disabled]="savingButtonEnable" (click)="submit(true)" class="text-button" color="primary" slot="end" fill="clear" size="small">Save</ion-button>
```
shawnm0705: `[disabled]="savingButtonEnable"` it's either a wrong variable name or a wrong logic... which one do you choose :)

shawnm0705: And since the condition of `ngIf` for the text and button are the same, we can put the condition in a `<ng-container>` wrapper

src/app/assessment/assessment.component.html
```diff
@@ -53,8 +55,11 @@ <h3>{{ group.name }}</h3>
                 [doReview]="doReview"
                 [submission]="(submission.answers && submission.answers[question.id]) ? submission.answers[question.id] : {}"
                 [review]="(review.answers && review.answers[question.id]) ? review.answers[question.id] : {}"
+                [reviewStatus]="review.status"
+                [submissionStatus]= "submission.status"
                 [formControlName]="'q-' + question.id"
                 [control]="questionsForm.controls['q-' + question.id]"
+                (saveProgress)="submit($event)"
```
shawnm0705: Since `$event` will always return `true`, we can just put `true` here instead of `$event`. Same as below

src/app/assessment/assessment.component.ts
```diff
@@ -129,12 +138,20 @@ export class AssessmentComponent extends RouterEnter {
         this.submission = result.submission;
         this.loadingSubmission = false;
         // this page is for doing assessment if submission is empty
```
shawnm0705: Update this comment

src/app/assessment/assessment.component.ts
```diff
@@ -192,35 +209,54 @@ export class AssessmentComponent extends RouterEnter {
...
 
-  submit() {
-    this.submitting = true;
+
+
+  submit(saveInProgress: boolean) {
+    if ( saveInProgress ) {
+      this.saving = true;
+      this.savingMessage = 'Saving...';
+      if (!this.savingButtonEnable) {
```
shawnm0705: Don't need to check it, just disable the saving button when it is saving

src/app/assessment/assessment.component.ts
```diff
@@ -192,35 +209,54 @@ export class AssessmentComponent extends RouterEnter {
...
       assessment = {
         id: this.id,
-        context_id: this.contextId
+        context_id: this.contextId,
+        in_progress: false
       };
+      if (saveInProgress) {
+        assessment.in_progress = true;
+      }
+      console.log('this.questionsForm.value', this.questionsForm.value);
```
shawnm0705: Remove this

src/app/assessment/assessment.component.ts
```diff
@@ -192,35 +209,54 @@ export class AssessmentComponent extends RouterEnter {
...
-        });
+        if (value) {
+          questionId = +key.replace('q-', '');
+          answers.push({
+            assessment_question_id: questionId,
+            answer: value
+          });
+        }
         // unset the required questions object
         if (requiredQuestions[questionId] && value) {
```
shawnm0705: This should be inside the above `if` condition

src/app/assessment/assessment.component.ts
```diff
@@ -192,35 +209,54 @@ export class AssessmentComponent extends RouterEnter {
...
         }
       });
-      // check if all required questions have answer
-      if (!this.utils.isEmpty(requiredQuestions)) {
-        this.submitting = false;
-        // display a pop up if required question not answered
-        return this.notificationService.popUp('shortMessage', {message: 'Required question answer missing!'});
+      // check if all required questions have answer when assessment done
+      if (!saveInProgress) {
+        if (!this.utils.isEmpty(requiredQuestions)) {
```
shawnm0705: Combine these two `if` together

src/app/assessment/assessment.component.ts
```diff
@@ -239,38 +279,52 @@ export class AssessmentComponent extends RouterEnter {
         }
       });
     }
-
     // save the submission/feedback
-    this.assessmentService.saveAnswers(assessment, answers, this.action)
-      .subscribe(result => {
+    console.log('saveAnswers', assessment, answers, this.action, this.submission.id);
```
shawnm0705: Remove this

src/app/assessment/assessment.component.ts
```diff
@@ -239,38 +279,52 @@ export class AssessmentComponent extends RouterEnter {
...
+    this.assessmentService.saveAnswers(assessment, answers, this.action, this.submission.id ? this.submission.id : undefined).subscribe(
```
shawnm0705: Actually we can just pass `this.submission.id` here, it will be 0 if there's no submission.

src/app/assessment/assessment.component.html
shawnm0705: And actually we don't event need `saving` this variable at all

src/app/questions/file/file.component.ts
```diff
@@ -113,4 +125,30 @@ export class FileComponent implements ControlValueAccessor, OnInit {
 
   }
 
+  // adding save values to from control
+  private _showSavedAnswers() {
+    if ((this.reviewStatus === 'in progress') && (this.doReview)) {
+      if (!this.innerValue) {
```
shawnm0705: At this time, the `innerValue` is not initialised for sure, so don't need this condition check
Same for all other files

src/app/questions/file/file.component.ts
```diff
@@ -113,4 +125,30 @@ export class FileComponent implements ControlValueAccessor, OnInit {
...
+      if (this.review.comment) {
+        this.innerValue.comment = this.review.comment;
+        this.comment = this.review.comment;
+      }
+      if (this.review.answer) {
+        this.innerValue.answer = this.review.answer;
+      }
+    }
+    if ((this.submissionStatus === 'in progress') && (this.doAssessment)) {
+      if (!this.innerValue) {
```
shawnm0705: Same here

src/app/questions/file/file.component.ts
```diff
@@ -113,4 +125,30 @@ export class FileComponent implements ControlValueAccessor, OnInit {
 
   }
 
+  // adding save values to from control
+  private _showSavedAnswers() {
+    if ((this.reviewStatus === 'in progress') && (this.doReview)) {
+      if (!this.innerValue) {
+        this.innerValue = {
+          answer: [],
```
shawnm0705: `answer` is not an array, it is an object

src/app/questions/file/file.component.ts
```diff
@@ -113,4 +125,30 @@ export class FileComponent implements ControlValueAccessor, OnInit {
...
+        this.innerValue.comment = this.review.comment;
+        this.comment = this.review.comment;
+      }
+      if (this.review.answer) {
+        this.innerValue.answer = this.review.answer;
+      }
+    }
+    if ((this.submissionStatus === 'in progress') && (this.doAssessment)) {
+      if (!this.innerValue) {
+        this.innerValue = [];
```
shawnm0705: Same here

src/app/questions/file/file.component.ts
```diff
@@ -113,4 +125,30 @@ export class FileComponent implements ControlValueAccessor, OnInit {
...
+      }
+      this.innerValue = this.submission.answer;
```
shawnm0705: Since we are assigning value to `innerValue` here, don't need the initialisation

src/app/questions/multiple/multiple.component.ts
```diff
@@ -101,5 +112,36 @@ export class MultipleComponent implements ControlValueAccessor {
...
+            answer: [],
+            comment: ''
+          };
+        }
+        this.innerValue.comment = this.review.comment;
+        this.comment = this.review.comment;
+      }
+      if (this.review.answer) {
+        if (!this.innerValue) {
+          this.innerValue = {
```
shawnm0705: Don't initialise it twice

src/app/questions/file/file.component.ts
```diff
@@ -113,4 +125,30 @@ export class FileComponent implements ControlValueAccessor, OnInit {
...
+          answer: [],
+          comment: ''
+        };
+      }
+      if (this.review.comment) {
+        this.innerValue.comment = this.review.comment;
+        this.comment = this.review.comment;
+      }
+      if (this.review.answer) {
+        this.innerValue.answer = this.review.answer;
```
shawnm0705: For both `comment` and `answer`, don't need to check if it is empty or not, we can just simply assign `this.review.answer` to `this.innerValue.answer` same for `comment`

Same for other files

.vscode/launch.json
sasangachathumal: i will delete it.

src/app/assessment/assessment.component.html
sasangachathumal: yes we don't need saving variable.

shawnm0705:
@sasangachathumal Looks good 👍

PR #322, opened 2019-03-08
[AV2-226]interaction on tapping

shawnm0705:


trtshen:
looks good! @shawnm0705

PR #346, opened 2019-03-29
[AV2-292]Event single booking feature

shawnm0705:


src/app/event-detail/event-detail.component.ts
```diff
@@ -67,6 +71,24 @@ export class EventDetailComponent {
...
+            {
+              text: 'OK',
+              role: 'cancel'
+            }
+          ]
+        });
+        // update the event list & activity detail page
+        this.utils.broadcastEvent('update-event', null);
+      }
+    });
```
trtshen: hey @shawnm0705 just wondering, do we need to also handle unsuccessful booking? just in case the booking unexpectedly failed?
not compulsory, but worth considering adding a simple feedback if something unexpected go bad.

shawnm0705: @trtshen Good point, changes have been committed, please review again

trtshen: thanks @shawnm0705

PR #347, opened 2019-04-03
[AV2-291] feature/custom-branding-login-screen

trtshen:


trtshen:
hi @shawnm0705 i am not sure what is the standard `experience/config` API data format, so currently I'm using:
- themeColor: for custom color (primary color)
- logo: for custom logo

Please do let me know if there are different values i should use instead of those, thanks!

src/app/auth/auth.module.ts
```diff
@@ -16,14 +16,16 @@ import { AuthForgotPasswordComponent } from './auth-forgot-password/auth-forgot-
...
 
 @NgModule({
   imports: [
     IonicModule,
     CommonModule,
     FormsModule,
     ReactiveFormsModule,
-    AuthRoutingModule
+    AuthRoutingModule,
+    SharedModule,
```
shawnm0705: Since we've used `SharedModule` here, we can remove `IonicModule, CommModule, FormsModule`

src/app/auth/auth-routing.module.ts
```diff
@@ -8,6 +8,7 @@ import { AuthRegistrationComponent } from './auth-registration/auth-registration
 import { AuthResetPasswordComponent } from './auth-reset-password/auth-reset-password.component';
 import { AuthDirectLoginComponent } from './auth-direct-login/auth-direct-login.component';
 import { UnauthorizedGuard } from './unauthorized.guard';
+import { AuthResolverService } from './auth-resolver.service';
```
shawnm0705: This has been removed already

src/app/app.component.ts
```diff
@@ -63,4 +81,11 @@ export class AppComponent implements OnInit {
     });
   }
 
+  getExpConfig(): Observable<any> {
+    let domain = window.location.hostname;
+    domain = (domain.indexOf('127.0.0.1') !== -1 || domain.indexOf('localhost') !== -1) ? 'appdev.practera.com' : domain;
```
shawnm0705: We don't need to force using `appdev.practera.com`, we can just use `localhost` or `127.0.0.1`. It's actually better for testing

PR #350, opened 2019-04-08
[AV2-252] Profile pictures on settings page

sasangachathumal:
added new css and html element for show profile image.

### DIFF
--- a/src/app/settings/settings.component.ts
+++ b/src/app/settings/settings.component.ts
@@ -19,7 +19,9 @@ export class SettingsComponent extends RouterEnter {
   routeUrl = '/app/settings';
   profile = {
     contactNumber: '',
-    email: ''
+    email: '',
+    image: '',
+    name: ''
   };
   currentProgramName = '';
   // default country model
@@ -48,6 +50,8 @@ export class SettingsComponent extends RouterEnter {
     // get contact number and email from local storage
     this.profile.email = this.storage.getUser().email;
     this.profile.contactNumber = this.storage.getUser().contactNumber;
+    this.profile.image = this.storage.getUser().image;
+    this.profile.name = this.storage.getUser().name;
     // also get program name
     this.currentProgramName = this.storage.getUser().programName;
     // if user has the contact number
--- a/src/app/settings/settings.component.ts
+++ b/src/app/settings/settings.component.ts
@@ -7,6 +7,7 @@ import { UtilsService, ContactNumberFormat } from '@services/utils.service';
 import { NotificationService } from '@shared/notification/notification.service';
 import { environment } from '../../environments/environment.prod';
 import { RouterEnter } from '@services/router-enter.service';
+import { FilestackService } from '@shared/filestack/filestack.service';
 
 @Component({
   selector: 'app-settings',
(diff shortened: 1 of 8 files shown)


src/app/settings/settings.component.ts
```diff
@@ -239,4 +247,36 @@ export class SettingsComponent extends RouterEnter {
     return true;
   }
 
+  async uploadProfileImage() {
+    const s3Config = this.filestackService.getS3Config('image');
+    const pickerOptions = {
```
shawnm0705: Do not need repeat all these, re-use filestack component, maybe just pass an additional parameter for the layout of the button

src/app/settings/settings.component.ts
```diff
@@ -31,6 +34,9 @@ export class SettingsComponent extends RouterEnter {
   helpline = 'help@practera.com';
 
   termsUrl = 'https://images.practera.com/terms_and_conditions/practera_terms_conditions.pdf';
+  // controll profile image updating
+  imageUpdating = false;
+  acceptFileTypes = ['.jpeg', '.jpg', '.png'];
```
shawnm0705: Don't use these file types, use `this.filestackService.getFileTypes(type)` instead

src/app/settings/settings.component.ts
```diff
@@ -48,6 +55,8 @@ export class SettingsComponent extends RouterEnter {
     // get contact number and email from local storage
     this.profile.email = this.storage.getUser().email;
     this.profile.contactNumber = this.storage.getUser().contactNumber;
+    this.profile.image = this.storage.getUser().image;
```
shawnm0705: Put a path of default image if no data passed in

src/app/settings/settings.component.ts
```diff
@@ -239,4 +248,27 @@ export class SettingsComponent extends RouterEnter {
...
+  async uploadProfileImage(file, type = null) {
+    if (file.success) {
+      this.imageUpdating = true;
+      this.settingService.updateProfileImage({
+        image: file.data.url
+        }).subscribe(
+        success => {
+          this.imageUpdating = false;
+          this.profile.image = file.data.url;
+          const casheUser = this.storage.getUser();
```
shawnm0705: Don't need to get the whole user object, it is handled inside the function already, check how other places use `setUser()`

src/app/shared/filestack/filestack.component.ts
```diff
@@ -4,12 +4,14 @@ import { FilestackService } from './filestack.service';
...
   selector: 'file-stack',
-  templateUrl: 'filestack.component.html'
+  templateUrl: 'filestack.component.html',
+  styleUrls: ['filestack.component.scss']
 })
 export class FilestackComponent implements OnInit {
   @Input() accept: any;
   @Input() fileType: string;
   @Output() complete: EventEmitter<any> = new EventEmitter();
+  @Input() profileImage?: boolean;
```
shawnm0705: Instead of using `profileImage` as the parameter, it'll be better to make it more generic to other cases in the future.
Call it `type` instead, and use `'profileImage'` as the value, and use switch/case in html for different cases. "Upload" button by default

src/app/settings/settings.component.ts
sasangachathumal: zaw done some changes, back end return image url always. if there are no image back end will send gravator url with default image added. so no need to put that from front end.

sasangachathumal: https://my.practera.com/img/user-512.png this url is the default image url this always added to image url back end returns.

PR #357, opened 2019-04-11
[AV2-298] Bugfix/large-attachment-preview

trtshen:


src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -477,8 +477,8 @@ export class ChatRoomComponent extends RouterEnter {
     );
   }
 
-  previewFile(file) {
-    this.filestackService.previewFile(file);
+  async previewFile(file) {
```
trtshen: hi @shawnm0705 I'm adding `async` to the code to handle Promise.
What do you think? For the sake fo consistency, this is likely breaking the rule, but in term of cleaner code, adding `async & await` is better option.

**Why `async & await` is better than Observable in such case?**

Because our original code was:
```javascript
// before
this.filestackService.previewFile(file);

// after adding asynchronies:
await this.filestackService.previewFile(file);

// otherwise, Observable way
this.filestackService.previewFile(file).subscribe(() => {
    // more logic should be added here
});
```

PR #360, opened 2019-04-17
[AV2-253] Profile picture on chat.

sasangachathumal:
- removed text avatar from chat list and chat room.
- remove colors from variables.scss and class from scss files.
- update chat service to remove unwanted methods.
- update html files to show profile image.
- update component ts files by removing unwanted methods.


sasangachathumal:
i have ask from anu and also, chaw and sanjaya. the told to keep our current chat ui. because if we add that round avatar to every message it not look good and if we need to add that we need to chat chat UI to something like slack chat. (adding sender name, not using rounded avatar). so that will be a separate chat UI. in this story we only need to add profile image to current UI. if we need to change UI need to do that in separate story and need to discuss with anu, beenish and sanja, easa.

PR #363, opened 2019-04-23
[AV2-293] bugfix/compulsory-question

trtshen:
- constant is immutable
- added FormControll flag at submit button

### DIFF
--- a/src/app/assessment/assessment.component.ts
+++ b/src/app/assessment/assessment.component.ts
@@ -57,7 +57,6 @@ export class AssessmentComponent extends RouterEnter {
   savingMessage = '';
   saving: boolean;
   fromPage = '';
-  canSubmit = false;
 
   constructor (
     public router: Router,
@@ -206,9 +205,15 @@ export class AssessmentComponent extends RouterEnter {
     return this.router.navigate(['app', 'home']);
   }
 
-  // form an object of required questions
-  getRequiredQuestions() {
+  /**
+   * @name compulsoryQuestionsAnswered
+   * @description to check if every compulsory question has been answered
+   * @param {Object[]} answers a list of answer object (in submission-based format)
+   */
+  compulsoryQuestionsAnswered(answers) {
+    const result = [];
     const requiredQuestions = {};
+    let missing = [];
     this.assessment.groups.forEach(group => {
       group.questions.forEach(question => {
         if (question.isRequired) {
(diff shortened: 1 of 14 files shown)


src/app/assessment/assessment.component.ts
```diff
@@ -57,6 +57,7 @@ export class AssessmentComponent extends RouterEnter {
   savingMessage = '';
   saving: boolean;
   fromPage = '';
+  canSubmit = false;
```
shawnm0705: This is not used anywhere

src/app/assessment/assessment.component.ts
```diff
@@ -218,6 +219,24 @@ export class AssessmentComponent extends RouterEnter {
     return requiredQuestions;
   }
 
+  getUnansweredCompulsoryQuestion(answers) {
+    const result = [];
+    const missing = [];
+    const required = this.getRequiredQuestions();
```
shawnm0705: Since we are using `getRequiredQuestions()` only here, we can just move the logic to loop through question from `getRequiredQuestions()` to here instead, so that we don't need to loop twice.

src/app/assessment/assessment.component.ts
```diff
@@ -218,6 +219,24 @@ export class AssessmentComponent extends RouterEnter {
...
+
+    this.utils.each(required, (question, questionId) => {
+      if (missing.length === 0) {
+        this.utils.each(answers, answer => {
```
shawnm0705: This can be replaced with a `filter()` function

trtshen: hey @shawnm0705 i dont think `filter()` works, i need to keep the datatype consistent (all should be `number`). As I know, by using `filter()` will just *filter* out data, but not modifying data. I need to [modify data into different format](https://github.com/intersective/practera-app-v2/pull/363/files/f39ee848eccd38719cadb7f208f84310482769c0#diff-0f829712a503e19c8721ec1a332df72aR231) for comparison purpose.

shawnm0705: @trtshen I mean use `answers.filter()` to get the `answer` we need, then we can still `missing.push({ assessment_question_id: +questionId, answer: answer.answer });`
```
const answer = answers.filter(answer => {
  return answer.assessment_question_id === +questionId;
});
if (this.utils.isEmpty(answer.answer)) {
  missing.push({ assessment_question_id: +questionId, answer: answer.answer });
}
```

src/app/assessment/assessment.component.ts
```diff
@@ -205,17 +205,34 @@ export class AssessmentComponent extends RouterEnter {
...
       group.questions.forEach(question => {
         if (question.isRequired) {
           requiredQuestions[question.id] = true;
         }
       });
     });
-    return requiredQuestions;
+
+    this.utils.each(requiredQuestions, (question, questionId) => {
+      if (missing.length === 0) {
```
shawnm0705: What I mean from last comment is to put this logic inside the above loop, so that we don't need to loop `requiredQuestions` again. We don't actually need `requiredQuestions` this variable
```
this.assessment.groups.forEach(group => {
  group.questions.forEach(question => {
    if (question.isRequired) {
      if (missing.length === 0) {
   .....
    }
  });
});
```

PR #373, opened 2019-05-06
Bugfix/circle and event api handling

trtshen:


jazzmind:
made a minor change because I set loadingEvents default to false in trial branch so changed it back to true.

PR #377, opened 2019-05-07
[AV2-308] upgrade to latest ionic

trtshen:


trtshen:
dependencies version follow latest ionic build, but still `tar` package having security issue which is sub-dependency used by package `@angular-devkit/build-angular`.

PR #380, opened 2019-05-14
Feature/av2 69/due dates on the app

sasangachathumal:
Jira Story - https://intersective.atlassian.net/browse/AV2-69
This is about showing due dates of assessments in assessment page and activity page. In activity page if due date is today we added a pulsing animation to icon of assessment.


src/app/services/utils.service.ts
```diff
@@ -205,4 +212,64 @@ export class UtilsService {
...
+   */
+  comparerDate(dateString: string, comparedDateString?: string) {
+    const date = new Date(dateString + 'Z');
+    let comparedDate = new Date();
+    if (comparedDateString) {
+      comparedDate = new Date(comparedDateString + 'Z');
+    }
+    // Calculate the difference and return
+    return Math.floor(
+      (Date.UTC(comparedDate.getFullYear(), comparedDate.getMonth(), comparedDate.getDate()) -
```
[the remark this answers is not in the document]
shawnm0705: Also, usually, for a function that compare/ calculate different between two elements, we use the first element to minus the second one. 
i.e. timeComparer(a, b) === -1 means a is before b
We should keep the same convention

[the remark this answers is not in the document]
shawnm0705: Actually, if we re-use `utcToLocal()` in the above function, we don't need to create this function anymore, we can just re-use `timeComparer()`

src/app/activity/activity.service.ts
```diff
@@ -111,7 +114,11 @@ export class ActivityService {
             type: 'Assessment',
             contextId: contextIds[element[element.model].id] || 0,
             loadingStatus: true,
-            isForTeam: element[element.model].is_team
+            isForTeam: element[element.model].is_team,
+            // @TODO need to update after api ready
```
shawnm0705: Do we still need this comment?

src/app/activity/activity.service.ts
```diff
@@ -111,7 +114,11 @@ export class ActivityService {
...
+            dueDate: this.utils.validateDueDates(element[element.model].deadline),
```
shawnm0705: Here, as an attribute of `Task`, `dueDate` should be just a date string, instead of how we display it.
We could define how we display it in html file

src/app/activity/activity.service.ts
```diff
@@ -111,7 +114,11 @@ export class ActivityService {
...
+            isOverdue: this.utils.validateDueDates(element[element.model].deadline).indexOf('Overdue') !== -1 ? true : false,
```
shawnm0705: `isOverdue` can be checked from `this.utils.timeComparer()`, better to not check the specific string 'Overdue'

src/app/activity/activity.service.ts
```diff
@@ -25,6 +25,9 @@ export interface Task {
   feedbackReviewed?: boolean;
   loadingStatus?: boolean;
   isForTeam?: boolean;
+  dueDate?: string;
+  isOverdue?: boolean;
+  iconClass?: string;
```
shawnm0705: change this to `isDueToday?: boolean`
We shouldn't put html class name inside files other than HTML/CSS file

src/app/assessment/assessment.component.html
```diff
@@ -13,7 +13,10 @@ <h1>{{ assessment.name }}</h1>
   <ion-text color="medium" text-left>
     <app-description *ngIf="assessment.description" [content]="assessment.description"></app-description>
   </ion-text>
-  <div class="review-submitter" *ngIf="!doAssessment" text-left>
+  <p [class]="assessment.isOverdue ? 'due-date assessment over' : 'due-date assessment'" *ngIf="doAssessment" text-center>
+    {{assessment.dueDate}}
+  </p>
+  <div class="review-submitter" *ngIf="!doAssessment && doReview" text-left>
```
shawnm0705: Why do we add `doReview` ?

src/app/assessment/assessment.service.ts
```diff
@@ -126,6 +128,8 @@ export class AssessmentService {
       name: thisAssessment.Assessment.name,
       description: thisAssessment.Assessment.description,
       isForTeam: thisAssessment.Assessment.is_team,
+      // @TODO need to update after api ready
```
shawnm0705: comment

src/app/assessment/assessment.component.ts
```diff
@@ -113,6 +117,7 @@ export class AssessmentComponent extends RouterEnter {
     this.assessmentService.getAssessment(this.id, this.action)
       .subscribe(assessment => {
         this.assessment = assessment;
+        this.assessment.isOverdue = assessment.dueDate.indexOf('Overdue') !== -1 ? true : false;
```
shawnm0705: move this to the service file

src/app/assessment/assessment.service.ts
```diff
@@ -126,6 +128,8 @@ export class AssessmentService {
...
+      dueDate: this.utils.validateDueDates(thisAssessment.Assessment.deadline, true),
```
shawnm0705: Same with the problem above

src/styles.scss
```diff
@@ -207,3 +207,53 @@ body {
...
+      vertical-align: bottom;
+    }
+  }
+  // use to show overdue dates
+  &.over {
+    color: var(--practera-ovderdue-color);
+  }
+}
+
+.pulseing {
```
shawnm0705: `pulseing` => `pulsing`

src/app/services/utils.service.ts
```diff
@@ -178,23 +170,56 @@ export class UtilsService {
...
+      if (date.getDate() === today.getDate() - 1) {
+        formattedDate = 'Yesterday';
+      }
+      if (date.getDate() === today.getDate()) {
+        formattedDate = 'Today';
+      }
+      if (date.getDate() === today.getDate() + 1) {
+        formattedDate = 'Tomorrow';
+      }
+      if (isDuedate && date.getDate() > today.getDate() + 1) {
```
shawnm0705: This is not necessary, we can use `short` for the month for due date as well

src/app/services/utils.service.ts
```diff
@@ -178,23 +170,56 @@ export class UtilsService {
...
+        formattedDate = new Intl.DateTimeFormat('en-GB', {
+          month: 'numeric',
           day: 'numeric',
           year: 'numeric'
         }).format(date);
+      }
     }
+
+    if (isDuedate) {
```
shawnm0705: This is not necessary, in `utcToLocal() : default` it should be
`return formattedDate + ' ' + formattedTime`

src/app/services/utils.service.ts
```diff
@@ -178,23 +170,56 @@ export class UtilsService {
...
+        hour12: true,
+        hour: 'numeric',
+        minute: 'numeric'
+      }).format(date);
+    }
+    return formattedDate;
   }
 
-  timeComparer(timeString: string, comparedString?: string) {
+  timeComparer(timeString: string, comparedString?: string, dueToday?: boolean) {
```
shawnm0705: this is `utils.service` so everything should be generic, `dueToday` is not a generic term here.
So the logic of this function is to compare two times. Instead of `dueToday`, we can use something like `compareDate?: boolean` to let this function compare the date difference instead of time difference. That will give us the information we need.

src/app/services/utils.service.ts
```diff
@@ -205,4 +230,21 @@ export class UtilsService {
...
+  /**
+   * This method check due dates of assessment or activity.
+   * - Check due date is today, tomorrow, upcoming date or overdue date.
+   * - If due date is upcoming one this will returns 'Due (date)' ex: 'Due 06-30-2019'.
+   * - If due date is overdue one this will returns 'Overdue (date)' ex: 'Overdue 01-10-2019'.
+   * - If due date is today this will return 'Due Today'.
+   * - If due date is tomorrow this will return 'Due Tomorrow'.
+   * @param dueDate - due date of assessment or activity.
+   */
+  dueDateFormatter(dueDate: string) {
```
shawnm0705: This shouldn't be in `utils.service`, put it in `shared.service`

src/app/activity/activity.service.ts
```diff
@@ -224,6 +230,9 @@ export class ActivityService {
         task.status = thisSubmission.AssessmentSubmission.status;
         break;
     }
+    if (task.status === 'done') {
```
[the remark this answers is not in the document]
sasangachathumal: if task status done not showing the `pulsing` animation.

src/app/activity/activity.component.html
```diff
@@ -35,11 +35,13 @@ <h1>{{ activity.name }}</h1>
             <ion-icon name="list-box" slot="start" color="primary"></ion-icon>
           </ng-container>
           <ng-container *ngSwitchCase="'Assessment'">
-            <ion-icon name="clipboard" slot="start" color="primary"></ion-icon>
+            <ion-icon name="clipboard" slot="start" [class]="(task.isDueToday && task.status != 'done') ? 'pulsing' : 'ng-star-inserted ion-color ion-color-primary hydrated'" color="primary"></ion-icon>
```
shawnm0705: `task.status != 'done'` is not enough, it is the same with below `(task.status === '' || task.status === 'in progress')`

src/app/services/utils.service.ts
```diff
@@ -178,23 +170,45 @@ export class UtilsService {
...
       default:
-        return new Intl.DateTimeFormat('en-GB', {
-          hour12: true,
-          hour: 'numeric',
-          minute: 'numeric',
-          month: 'short',
-          day: 'numeric',
-          year: 'numeric'
-        }).format(date);
+      return this.dateFormatter(date) + ' ' + new Intl.DateTimeFormat('en-GB', {
```
shawnm0705: Do not repeat it, 
`const formattedTime = new Intl.DateTimeFormat('en-GB', ...`
and use it in both `case 'time'` and `default`

src/app/services/utils.service.ts
```diff
@@ -178,23 +170,45 @@ export class UtilsService {
...
   }
 
-  timeComparer(timeString: string, comparedString?: string) {
+  timeComparer(timeString: string, comparedString?: string, compareDate?: boolean) {
     const time = new Date(timeString + 'Z');
     let compared = new Date();
     if (comparedString) {
       compared = new Date(comparedString + 'Z');
     }
+    if (compareDate && (time.getDate() === compared.getDate() && time.getTime() >= compared.getTime())) {
```
shawnm0705: `time.getTime() >= compared.getTime()` is not necessary.
Also, we need to compare year and month here to make sure it's the same day

PR #533, opened 2019-09-04
[AV2-432] Alignment issue on the home page

sasangachathumal:
there is 3 main issues.
- There is alignment issue on the home page, where there is no padding between 'Current activity' and the activity card.
- There is no padding between milestone name and activity card on the milestone page.
- On the milestone page, there is inconsistent padding between the milestone name and activity card.
JIra story - https://intersective.atlassian.net/browse/AV2-432

### DIFF
--- a/src/app/project/project.component.html
+++ b/src/app/project/project.component.html
@@ -38,17 +38,14 @@
         </ng-container>
 
         <ng-template #actualMilestoneTpl>
-          <div class="milestone ion-padding-top">
+          <div class="milestone" [ngClass]="{'lock': milestone.isLocked}">
             <div>
                 <ion-icon *ngIf="milestone.isLocked" name="lock" color="medium" class="icon-lock milestone-lock"></ion-icon>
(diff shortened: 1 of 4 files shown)


sasangachathumal:
Fix i have done.

Added margin to card top 16px and bottom 16px. because we need to keep 32px space between two cards.

this change effect to fix.
        1. space issue between card and ‘current activity’ text in dashboard.
        2. inconsistent padding between the milestone name and activity card.

Added padding to milestone in top 16px. And for lock milestone added 16 px in bottom. Because if there are activities in milestone it will adding 16px space in the end of milestone. When it is locked milestone there are no cards inside.

this change effect to fix.
        1. Padding issue between tow lock milestones.
        2. Correcting space between last activity card and milestone name.

Added padding to card according to design. Top and bottom 12px, right and left 16px.

this change effect to fix.
        1. Fit all activity description inside card.
        2. style correcting according to design.

Added some space to header icons. Tick and lock.

this change effect to fix.
        1. Icon alignment issue in header.

PR #566, opened 2019-09-25
[AV2-443] Autosave not available

sasangachathumal:
Jira story - https://intersective.atlassian.net/browse/AV2-443


Barry-Chen-Intersective:
https://qareport.practeraco.de/appv2_coverage/report_13-10-01-25-09-2019/result.html

src/app/assessment/assessment.component.ts
```diff
@@ -455,11 +455,32 @@ export class AssessmentComponent extends RouterEnter {
...
+     */
     if (saveInProgress) {
-      this.savingMessage = 'Saving...';
-      this.savingButtonDisabled = true;
+      if (isManualSave || !this.saving) {
+        this.savingMessage = 'Saving...';
+        this.saving = true;
+        this.savingButtonDisabled = true;
+      } else {
+        return;
```
shawnm0705: If `isManualSave = false` , it will return directly here. 
That means we will never trigger auto save?

PR #567, opened 2019-09-26
[AV2-447] Activity List / Detail Accessibility Fix

sasangachathumal:
Jira Story - https://intersective.atlassian.net/browse/AV2-447

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -65,6 +65,13 @@ body {
     max-width: 80%;
   }
 
+  .task-name {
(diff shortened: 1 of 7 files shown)


src/styles.scss
```diff
@@ -65,6 +65,13 @@ body {
     max-width: 80%;
   }
 
+  .task-name {
```
shawnm0705: Why do we need this in the global `style.scss`? 
If it is only for one html, put that in the scss under the same folder

PR #618, opened 2019-11-21
[AV2-480] Desktop view - chat

sasangachathumal:
Jira story - https://intersective.atlassian.net/browse/AV2-480.
need to keep all functionality of two components (chat list and chat room) and load both components to one component to show on desktop view.
and need to add more things when showing the desktop view.


src/app/chat/chat-view/chat-view.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.participantsOnly = undefined;
+    this.teamId = undefined;
+    this.chatName = undefined;
+  }
+
+  goto(event) {
+    this.teamId = event.teamId;
+    this.teamMemberId = event.teamMemberId;
+    this.participantsOnly = event.participantsOnly;
+    this.chatName = event.chatName;
```
shawnm0705: what if event don't have chatName

src/app/chat/chat-view/chat-view.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    // navigate to the first chat
+    this.goto({
+      teamId: chats[0].team_id,
+      teamMemberId: chats[0].team_member_id,
+      participantsOnly: chats[0].participants_only,
+      chatName: chats[0].name
+    });
+  }
+
+  private _gotoChat() {
```
shawnm0705: add inline comment explaining what does this do

src/app/tabs/tabs.component.ts
```diff
@@ -54,6 +54,13 @@ export class TabsComponent extends RouterEnter {
         });
       });
     }
+    if (!this.utils.isMobile()) {
+      this.utils.getEvent('chat-bubble-update').subscribe(event => {
+        this.tabsService.getNoOfChats().subscribe(noOfChats => {
```
shawnm0705: no need to call api again to get the number of chats, it can be calculated here

src/app/chat/chat-list/chat-list.component.html
```diff
@@ -4,7 +4,7 @@
   </ion-toolbar>
 </ion-header>
 
-<ion-content color="light">
+<ion-content color="light" class="ion-content-absolute">
```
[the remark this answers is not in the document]
sasangachathumal: there is no class called `ion-header-absolute` but in `topic.component.html` you use `ion-toolbar-absolute`.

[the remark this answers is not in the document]
sasangachathumal: I have added `ion-toolbar-absolute` to the chat list and chat room components.

src/app/chat/chat-room/chat-room.component.html
```diff
@@ -1,20 +1,20 @@
 <ion-header mode="ios">
```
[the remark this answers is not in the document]
sasangachathumal: there is no class `ion-header-absolute` but we have `ion-toolbar-absolute`. so I added that.
https://github.com/intersective/practera-app-v2/blob/desktop-view/src/styles.scss#L7

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -120,13 +126,27 @@ export class ChatRoomComponent extends RouterEnter {
...
+    if (this.teamMemberId) {
+      this.selectedChat.team_member_id = this.teamMemberId;
+    } else if (Number(this.route.snapshot.paramMap.get('teamMemberId'))) {
       this.selectedChat.team_member_id = Number(this.route.snapshot.paramMap.get('teamMemberId'));
+    } else if (this.participantsOnly) {
+      this.selectedChat.is_team = true;
+      this.selectedChat.participants_only = this.participantsOnly;
     } else {
       this.selectedChat.is_team = true;
       this.selectedChat.participants_only = JSON.parse(this.route.snapshot.paramMap.get('participantsOnly'));
```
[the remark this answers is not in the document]
sasangachathumal: if we break into something like this.
`if (this.teamMemberId) {
      this.selectedChat.team_member_id = this.teamMemberId;
    } else {
      this.selectedChat.team_member_id = Number(this.route.snapshot.paramMap.get('teamMemberId'));
    }
    if (this.participantsOnly) {
      this.selectedChat.is_team = true;
      this.selectedChat.participants_only = this.participantsOnly;
    } else {
      this.selectedChat.is_team = true;
      this.selectedChat.participants_only = JSON.parse(this.route.snapshot.paramMap.get('participantsOnly'));
    }`
`this.selectedChat.is_team` will be always true. because last else statement call always.
to fix that if did something like this.
`if (this.teamMemberId) {
      this.selectedChat.team_member_id = this.teamMemberId;
      return;
    } else {
      this.selectedChat.team_member_id = Number(this.route.snapshot.paramMap.get('teamMemberId'));
      return;
    }
    if (this.participantsOnly) {
      this.selectedChat.is_team = true;
      this.selectedChat.participants_only = this.participantsOnly;
    } else {
      this.selectedChat.is_team = true;
      this.selectedChat.participants_only = JSON.parse(this.route.snapshot.paramMap.get('participantsOnly'));
    }`
it will not check `participantsOnly` because if team member id is not there it will go to else block and add undefiend to `this.selectedChat.team_member_id` and `return;`.

what I want to do is,
if we have team member id, get it from @Input parameter or rout parameter then not checking for participants Only.
if we didn't have team member id as any parameter, then only get participants Only from @Input parameter or rout parameter and make `this.selectedChat.is_team = true`.

src/app/tabs/tabs.component.ts
sasangachathumal: if we `event.readcount` to calculate. it will be the wrong count. because `readcount` is how many message read it does not always equal to unread message count it can be more than that. as `readcount` we got every message mark as read message-id Array length, we send every id to mark as read, even it's not an unread message. if we call API as we doing in other events we can get correct unread message count set it to tab.

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -196,9 +215,13 @@ export class ChatRoomComponent extends RouterEnter {
 
   private _getChatName() {
```
shawnm0705: The correct if/else for this function should be 
```
if it is team chat {
  get team name from service
  return 
}
if chat name is in variable {
  get chat name from variable
  return
}
if chat name is passed in as parameter {
  get chat name from parameter
  return
}
get the chat title from message list

File: src/app/chat/chat-view/chat-view.component.ts [hunk 8]
```diff
@@ -0,0 +1,75 @@
...
+    return {
+      teamId: this.teamId,
+      chatName: this.chatName,
+      teamMemberId: this.teamMemberId,
+      participantsOnly: this.participantsOnly
+    };
+  }
+
+  selectFirstChat(chats) {
+    if (this.teamId) {
```
shawnm0705: add some inline comment on this

src/app/chat/chat-list/chat-list.component.ts
```diff
@@ -82,43 +99,72 @@ export class ChatListComponent extends RouterEnter {
     }
   }
 
+  // force every navigation happen under radar of angular
+  private _navigate(direction) {
+    if (this.utils.isMobile()) {
+      // remove name from passing values
+      // this name is receiver name, we pass this to set to header in char room.
```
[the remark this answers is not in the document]
sasangachathumal: yes, after we change `_getChatName()` method logic to what you mention in another comment we no need to remove it. it will handle from `_getChatName()`.

PR #638, opened 2019-12-03
[AV2-472] Task/unit test for pusher service

trtshen:


trtshen:
@shawnm0705 the ngzone is causing pusher is no longer being watched by angular and tab badge number no longer updating even new pusher event is pushed to appv2. In order to make this works for protractor, we should figure out another way instead of just adding ngzone to pusher code.

PR #668, opened 2020-01-05
hotfix for IE/Edge text reversal on Windows

jazzmind:
@chaw this should fix the iE text reversal problem. Please test and deploy ASAP.


Mihai-P:
I changed the base to develop instead of master.

PR #706, opened 2020-01-22
[AV2-505] redirect user to registration page from direct link

sasangachathumal:
Jira story - https://intersective.atlassian.net/browse/AV2-505
### **I'm not sure these changes are the optimal solution, so please give me your suggestions to improve this. and I'm still didn't find a way to test this.**


src/app/auth/auth-direct-login/auth-direct-login.component.ts
```diff
@@ -55,7 +55,7 @@ export class AuthDirectLoginComponent implements OnInit {
   /**
    * Redirect user to a specific page if data is passed in, otherwise redirect to program switcher page
    */
-  private async _redirect(): Promise<boolean> {
+  private async _redirect(notRegister?: boolean) {
```
trtshen: @sasangachathumal it would be best for us to keep the return type for function.

trtshen: in case you not sure, it might be something like `Promise<boolean | string>`

src/app/switcher/switcher.service.ts
```diff
@@ -185,7 +185,12 @@ export class SwitcherService {
       this.utils.clearCache();
       if ((typeof environment.goMobile !== 'undefined' && environment.goMobile === false)
         || /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)) {
-        return ['app', 'home'];
+          if (this.storage.get('directLinkData')) {
```
shawnm0705: Since it's a route, it's better to name it `directLinkRoute` instead

src/app/auth/auth-direct-login/auth-direct-login.component.ts
shawnm0705: I think it's better to name this variable something like `redirectLater` or `delayedRedirect`. So that from the name we know we are going to redirect the user later.
Otherwise, we will forget what does "not register" mean to the function "redirect".

src/app/auth/auth-direct-login/auth-direct-login.component.ts
```diff
@@ -122,6 +125,7 @@ export class AuthDirectLoginComponent implements OnInit {
             if (!this.utils.isEmpty(res) && res.status === 'forbidden' && [
               'User is not registered'
             ].includes(res.data.message)) {
+              this.storage.set('directLinkData', this._redirect(true));
```
shawnm0705: The logic to set the redirect route into local storage should be part of the function `_redirect()`
So `_redirect()` will either save the route to local storage for later redirection or redirect directly.
Given that, we can create a function called `_saveOrRedirect(route: Array, save = false)`, so that we can call this in the switch/case branches inside `_redirect()`

src/app/auth/auth-direct-login/auth-direct-login.component.ts
```diff
@@ -66,66 +66,76 @@ export class AuthDirectLoginComponent implements OnInit {
...
-        return this.navigate(['app', 'home']);
+      return this._saveOrRedirect(['app', 'home'], redirectLater);
     }
-    return this.navigate(['app', 'home']);
+    return this._saveOrRedirect(['app', 'home'], redirectLater);
+  }
+
+  private _saveOrRedirect(route: Array<String | number>, save = false) {
+    if (save) {
+      return this.storage.set('directLinkRoute', JSON.stringify(route));
```
shawnm0705: No need to JSON.stringify when calling storage service set()

src/app/switcher/switcher.service.ts
```diff
@@ -185,7 +185,12 @@ export class SwitcherService {
       this.utils.clearCache();
       if ((typeof environment.goMobile !== 'undefined' && environment.goMobile === false)
         || /iPhone|iPad|iPod|Android/i.test(navigator.userAgent)) {
-        return ['app', 'home'];
+          if (this.storage.get('directLinkRoute')) {
+            const route = this.storage.get('directLinkRoute');
+            this.storage.remove('directLinkRoute');
+            return JSON.parse(route);
```
shawnm0705: No need to JSON.parse as well

PR #727, opened 2020-02-06
[AV2-570] Desktop View - achievements page

sasangachathumal:
**in this PR** 
update UI to fixed issues found in code review and done some code optimization


src/app/achievements/achievements.component.ts
```diff
@@ -36,11 +36,11 @@ export class AchievementsComponent extends RouterEnter {
       image: '',
       name: ''
     };
+    this.getUser();
```
shawnm0705: Don't need to create a function here, just assign the correct value to `this.userInfo` above

PR #730, opened 2020-02-07
[AV2-418] Feature/program-switcher-unit-test

trtshen:


src/app/switcher/switcher.service.ts
```diff
@@ -76,6 +76,7 @@ export class SwitcherService {
   }
 
   switchProgram(programObj: ProgramObj): Observable<any> {
+    console.log('test?');
```
shawnm0705: Shall we remove this?

trtshen: GOOD CATCH!

PR #753, opened 2020-02-20
Feature/av2 611/auto resize switcher image using filestack url parameters

sasangachathumal:
Jira - https://intersective.atlassian.net/browse/AV2-611
**In this PR fixed issue,**
a new participant enrolls into a new program, the URL in the settings page does not resize with parameters.

### DIFF
--- a/src/app/settings/settings.component.ts
+++ b/src/app/settings/settings.component.ts
@@ -35,6 +35,8 @@ export class SettingsComponent extends RouterEnter {
   // controll profile image updating
   imageUpdating = false;
   acceptFileTypes;
+  // card image CDN
+  cdn = 'https://cdn.filestackcontent.com/resize=fit:crop,width:';
 
   constructor (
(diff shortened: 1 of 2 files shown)


src/app/settings/settings.component.ts
```diff
@@ -61,10 +63,22 @@ export class SettingsComponent extends RouterEnter {
     this.acceptFileTypes = this.filestackService.getFileTypes('image');
     // also get program name
     this.currentProgramName = this.storage.getUser().programName;
-    this.currentProgramImage = this.storage.getUser().programImage;
+    this._getCurrentProgramImage();
```
shawnm0705: It's better to let `_getCurrentProgramImage()` return the image url and assign it to `this.currentProgramImage` inside this function.

sasangachathumal: ok. I have pushed an update.

PR #758, opened 2020-02-24
[AV2-601] WCAG color requirements

sasangachathumal:
Jira - https://intersective.atlassian.net/browse/AV2-601
In this story only fixing font colors. not font sizes.

### DIFF
--- a/src/theme/pfas/_colors.scss
+++ b/src/theme/pfas/_colors.scss
@@ -5,16 +5,32 @@
   color: #F5F6FA !important;
 }
 .gray-0 {
-  color: #ABAEBB !important;
+  color: rgba(0, 0, 0, 0.3) !important;
 }
 .gray-1 {
-  color: #A4A8B7 !important;
+  color: rgba(0, 0, 0, 0.4) !important;
 }
 .gray-2 {
-  color: #818496 !important;
+  color: rgba(0, 0, 0, 0.6) !important;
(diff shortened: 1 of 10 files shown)


sasangachathumal:
PR still not ready for code review

src/theme/pfas/_colors.scss
```diff
@@ -5,16 +5,24 @@
...
 .gray-3 {
-  color: #565B70 !important;
+  color: rgba(0, 0, 0, 0.8) !important;
+}
+.black {
+  color: black !important;
+}
+// use in activity, event components
+// for list item shared component for expired event title
+.color-gray-2 {
```
shawnm0705: no need anymore

sasangachathumal: @shawnm0705 In activity component and in events we pass this 
`[titleColor]="eventListService.isNotActionable(event) ? 'gray-2' : ''"`
and in list item component we put title color like this
`<p class="item-title subtitle-1" [ngClass]="'color-' + titleColor">{{ title }}</p>`

sasangachathumal: that's why i create a class color-gray-2

shawnm0705: Then we just need to put this class inside the css of list-item component only, it is not used anywhere else

PR #767, opened 2020-03-04
[AV2-414] Task/generic services unit test

trtshen:
https://intersective.atlassian.net/browse/AV2-414


trtshen:
> Can I ask why you imported SharedModule here? do we need that?

hi @ronakamani , the short answer, it's because GoMobileComponent itself is part of the GoMobileModule. And GoMobileModule need the SharedModule imported.

Full explanation:
From your question, I think you've gone through the code inside GoMobileComponent, and found that GoMobileComponent don't need SharedModule as dependency (imported into the constructor), which it might be the reason causing you to wonder why do we need that in SharedModule in the providers field in `TestBed.configureTestingModule()`.

In order to get GoMobileComponent to work, we need to setup base for the GoMobileComponent, in this case, it's GoMobileModule. Inside GoMobileModule, we imported SharedModule, this import tells GoMobileModule that our code need some elements from SharedModule.

One of the reused component from SharedModule, the `ContactNumberFormComponent`. In our GoMobileComponent's template code, you'll find that we are using `<app-contact-number-form></app-contact-number-form>` which it's belongs to `ContactNumberFormComponent` from SharedModule, that's the reason we need that imported in TestBed to setup a full exact test target as the original one.

ronakamani:
thanks Chaw

PR #783, opened 2020-03-12
Task/[AV2-579]/navigation-tabs-disappearing

ronakamani:


src/app/tabs/tabs.component.ts
```diff
@@ -65,12 +65,14 @@ export class TabsComponent extends RouterEnter {
         });
       });
     }
+    this.onEnter();
```
shawnm0705: As the onEnter() is invoked when ngOnInit() in `RouterEnter.service`, we can't re-invoke it here, otherwise, it is duplicated.
We need to find out why the ngOnInit() doesn't get triggered for this specific scenario

ronakamani: right. thanks @shawnm0705.

PR #812, opened 2020-03-31
Feature/av2 299/mov video files are not being transcoded

sasangachathumal:
Jira story - https://intersective.atlassian.net/browse/AV2-299

### DIFF
--- a/src/app/chat/chat-room/chat-room.component.html
+++ b/src/app/chat/chat-room/chat-room.component.html
@@ -127,7 +127,7 @@
           (ngModelChange)="typing()"
           appAutoresize="100"
           rows="1"
-          (ionFocus)="textFieldFocused = true; _scrollToBottom();"
+          (ionFocus)="textFieldFocused = true;"
         ></ion-textarea>
       </ion-col>
       <ng-container *ngIf="utils.isMobile() && !textFieldFocused">
--- a/src/app/chat/chat-room/chat-room.component.html
+++ b/src/app/chat/chat-room/chat-room.component.html
@@ -131,39 +131,11 @@
         ></ion-textarea>
       </ion-col>
       <ng-container *ngIf="utils.isMobile() && !textFieldFocused">
-        <ion-col class="ion-no-padding ion-text-center">
-          <ion-button (click)="attach('video')" fill="clear" class="ion-no-padding">
-            <ion-icon name="film-outline" slot="icon-only"></ion-icon>
-          </ion-button>
-        </ion-col>
-        <ion-col class="ion-no-padding ion-text-center">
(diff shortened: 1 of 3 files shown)


src/app/chat/chat-room/chat-room.component.html
```diff
@@ -137,9 +127,44 @@
...
+        <ion-col class="ion-no-padding ion-text-center">
+          <ion-button (click)="attach('image')" fill="clear" class="ion-no-padding">
+            <ion-icon name="images-outline" slot="icon-only"></ion-icon>
+          </ion-button>
+        </ion-col>
+        <ion-col class="ion-no-padding ion-text-center">
+          <ion-button (click)="attach('any')" fill="clear" class="ion-no-padding">
+            <ion-icon name="attach-outline" slot="icon-only"></ion-icon>
+          </ion-button>
+        </ion-col>
```
shawnm0705: Try to reuse this block of HTML with `ng-template` so that we are not duplicating codes

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -39,6 +39,7 @@ export class ChatRoomComponent extends RouterEnter {
   loadingMesageSend = false;
   isTyping = false;
   typingMessage: string;
+  textFieldFocused = false;
```
shawnm0705: Please add some comments on what does this do

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -39,6 +39,9 @@ export class ChatRoomComponent extends RouterEnter {
   loadingMesageSend = false;
   isTyping = false;
   typingMessage: string;
+  // this use to show/hide bottom section of text field which have attachment buttons and send button,
+  // when user typing text messages
+  textFieldFocused = false;
```
shawnm0705: If this flag is only used to show/hide the buttons, it's better to name it something like "hideAttachmentButtons" so that it is self-explained. Otherwise "textFieldFocused" is not a good name to explain the situation

src/app/chat/chat-room/chat-room.component.html
```diff
@@ -137,13 +127,38 @@
...
         <ion-button (click)="sendMessage()" type="submit" fill="clear" class="ion-no-padding">
           <ion-icon name="send-outline" slot="icon-only"></ion-icon>
         </ion-button>
       </ion-col>
     </ion-row>
   </ion-grid>
 </ion-footer>
+
+<ng-template #attachmentButtons let-offset="offset">
+  <ion-col offset="{{ offset }}" class="ion-no-padding ion-text-center">
```
shawnm0705: It's better to use `[offset]="offset"` instead

PR #834, opened 2020-04-19
Feature/av2 597/change the way we display practera videos

sasangachathumal:
Jira story - https://intersective.atlassian.net/browse/AV2-597

### DIFF
--- a/src/app/topic/topic.component.ts
+++ b/src/app/topic/topic.component.ts
@@ -61,7 +61,6 @@ export class TopicComponent extends RouterEnter {
     @Inject(DOCUMENT) private readonly document: Document
   ) {
     super(router);
-    // this.player = new Plyr('#player', {});
   }
 
   private _initialise() {
@@ -79,7 +78,6 @@ export class TopicComponent extends RouterEnter {
     this.isLoadingPreview = false;
     this.btnToggleTopicIsDone = false;
     this.askForMarkAsDone = false;
-    // console.log(new Plyr('#player', {}));
   }
 
   onEnter() {
@@ -96,7 +94,17 @@ export class TopicComponent extends RouterEnter {
     }
     this._getTopic();
     this._getTopicProgress();
-    setTimeout(() => this.player = new Plyr(this.document.getElementById('player'), {ratio: '16:9', autoplay: true}), 2000);
+    setTimeout(() => {
+      const players = Array.from(document.querySelectorAll('#video')).map(p => console.log('map element - ', p));
+      const videoEle = this.document.getElementById('video');
+      console.log('normal element - ', videoEle);
+      this.player = new Plyr(videoEle, {ratio: '16:9'});
+      videoEle.classList.add('topic-video');
+      if (!this.utils.isMobile()) {
+        videoEle.classList.add('desktop-view');
+      }
+    // tslint:disable-next-line:align
+    }, 2000);
     setTimeout(() => this.askForMarkAsDone = true, 15000);
   }
 
--- a/src/app/topic/topic.component.ts
+++ b/src/app/topic/topic.component.ts
@@ -43,7 +43,6 @@ export class TopicComponent extends RouterEnter {
   isLoadingPreview = false;
   askForMarkAsDone: boolean;
   redirecting = false;
-  player;
 
   constructor(
     private topicService: TopicService,
--- a/src/app/topic/topic.component.ts
+++ b/src/app/topic/topic.component.ts
@@ -94,15 +94,16 @@ export class TopicComponent extends RouterEnter {
     }
     this._getTopic();
     this._getTopicProgress();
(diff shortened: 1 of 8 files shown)


src/app/topic/topic.component.ts
```diff
@@ -90,6 +93,18 @@ export class TopicComponent extends RouterEnter {
     }
     this._getTopic();
     this._getTopicProgress();
+    // convert other brand video players to custom player.
+    setTimeout(() => {
+      Array.from(this.document.querySelectorAll('#video')).map(player => {
```
shawnm0705, line 98: we only have one `#video`, don't need to do a queryselectall

src/app/topic/topic.component.ts
```diff
@@ -90,6 +93,18 @@ export class TopicComponent extends RouterEnter {
...
+        // tslint:disable-next-line:no-unused-expression
+        new Plyr(player as HTMLElement, {ratio: '16:9'});
+        player.classList.add('topic-video');
```
shawnm0705, line 101: Why can't we just add these class names into the HTML directly?

src/app/topic/topic.component.ts
sasangachathumal: yes in our component we have only one element with id="video", but if you check story i have mention in there this also working with embedded videos. so we can't say only one video embedded. if video added from video URL html element in the code will show it but if video is embedded that html code shows after view loaded.

sasangachathumal: because we have embedded videos and also video URL that's why i use `queryselectall`. if their only one element or multiple element `queryselectall` can handle that.

src/app/topic/topic.component.ts
sasangachathumal: main reason is plyr plugin replacing all css classes when it render the player.
if we adding css class directly from html, we need to add that to this element `<div class="plyr__video-embed" id="video" [innerHtml]="iframeHtml"></div>` but when we create player from this line `new Plyr(player as HTMLElement, {ratio: '16:9'});` plugin is replacing all css class that element have. so our class will removed my plugin.
another reason is if we add css class in html we can't add then to embedded videos.

src/app/topic/topic.component.ts
sasangachathumal: i have test in practera web, we can create topics that have multiple embedded videos. and also we can create topic that have both embedded video and video URL.

shawnm0705: How do we make sure the embedded video has the same id `video` as in our component?

shawnm0705: If we have to deal with the embedded video, shall we consider using a smarter way to detect those videos? 
We can't guarantee that all videos have the `id="video"`. 
If we are dealing with only `<iframe>` and  `<video>` these two scenarios, we can use the CSS query selector to get those DOMs, and make sure we are only picking those DOMs inside the `<app-topic>`

sasangachathumal: we need to add that when we create topic. i had chat with anu about this. we need to inform this to programs team. when they embedded videos they need to add an extra `div` with `id` and `class`. i have include examples in the jira story. if they didn't add this extra `div` video will play on other branded players. what anu suggested is. let's try this first if programs team have difficulties to add this extra html code when embedding video we may need to change that.

shawnm0705: Also, if we have to do a `querySelectorAll`, it's better to do `querySelectorAll` against a class name instead of id. Something like `querySelectorAll('.topic-video')`. `id` of an HTML element is meant to be unique, while `class` is used to set the element under a "group".

shawnm0705: Also, we are not actually "mapping" an Array, we are just looping through it, so instead of doing an `Array.map()`, use `Array.forEach()` instead

sasangachathumal: yes we can use CSS query selector.

sasangachathumal: ok will change it to Array.forEach().

src/app/topic/topic.service.ts
```diff
@@ -73,6 +73,8 @@ export class TopicService {
     // to privent that we checking topic content is not equels to empty string.
     if (this.utils.has(thisTopic.Story, 'content') && !this.utils.isEmpty(thisTopic.Story.content)) {
       thisTopic.Story.content = thisTopic.Story.content.replace(/text-align: center;/gi, 'text-align: center; text-align: -webkit-center;');
+      thisTopic.Story.content = thisTopic.Story.content.replace(/(<iframe*)/g, '<div class="plyr__video-embed"><iframe').replace(/(<\/iframe>*)/g, '</iframe></div>');
```
shawnm0705, line 76: no need to put `*` there

PR #835, opened 2020-04-20
[AV2-640]/unassigned review msg to mentor

ronakamani:


src/app/assessment/assessment.component.ts
```diff
@@ -250,6 +250,21 @@ export class AssessmentComponent extends RouterEnter {
               ]
             });
           }
+          if (this.action === 'review' && !this.doReview) {
```
shawnm0705, line 253: This is not the correct condition for this situation, if action === 'review' and !this.doReview that means the user is seeing the review.

ronakamani: @shawnm0705 should I just add the role to the condition?

shawnm0705: It’s nothing about checking the mentor role or not

The scenario is that the mentor is not assigned to the review, but he goes to the review page

The correct logic is
- API should not return the review data
- Front-end should pop up toast message if there’s no review data

src/app/assessment/assessment.component.ts
```diff
@@ -331,6 +331,21 @@ export class AssessmentComponent extends RouterEnter {
         }
       },
       (error) => {
+        if (!this.review.status) {
```
shawnm0705, line 334: This is still not the right place to put these codes. 
I'm working on [AV2-646] which will change the code structure of this part, so I think it's better to hold this story for now and go back to it after [AV2-646] has been merged.

ronakamani: ok, sure. thanks @shawnm0705

shawnm0705:
The graphql implementation as been merged to develop branch so that we can re-start this story now

Barry-Chen-Intersective:
It need some changes to the develop branch. While we unassigns a mentor and access the review the direct link, the page display as an empty submission rather than a review.

PR #859, opened 2020-05-18
fix the issue

ronakamani:


src/app/shared/notification/achievement-pop-up/achievement-pop-up.component.html
```diff
@@ -22,7 +22,7 @@
       <app-description *ngIf="achievement.description" class="body-1 gray-2" [content]="achievement.description" [isInPopup]="true"></app-description>
     </ion-text>
 
-    <div class="div-after-img button-container ion-text-center">
+    <div class="div-after-img button-container text-center">
```
shawnm0705, line 25: Why remove it?

ronakamani: it wasn't align on chrome with ion-text-centre class

shawnm0705: @ronakamani and we don't have class named with 'text-center'. so if we don't need anymore, just remove the class totally

PR #922, opened 2020-07-27
[AV2-723]Refactor the use of graph ql

shawnm0705:


src/app/activity/activity.service.ts
```diff
@@ -58,37 +58,35 @@ export class ActivityService {
...
-          `id name description tasks{` +
-            `id name type isLocked isTeam deadline contextId status{` +
-              `status isLocked submitterName submitterImage` +
-            `}` +
-          `}` +
-        `}` +
-      `}"`)
-      .pipe(map(res => this._normaliseActivity(res.data)));
+    return this.request.graphQLQuery(
+      `query getActivity($id: Int!) {
```
trtshen, line 63: this is really nice, it is so much cleaner now!

PR #976, opened 2020-09-17
[AV2-753] bugfix/made-backward-compatible

trtshen:
this fix is part of the https://intersective.atlassian.net/browse/AV2-750 (to make hybrid alpha build work correctly)

### Context of this fix
I think there is recent changes that repeatedly trigger fast-feedback form popup in the older test program on AppV2 development environment.

The new change is also causing the new fast-feedback form modal get popup without Meta value [this Endpoint](https://sandbox.practera.com/api/v2/observation/slider/list.json), so then user can never submit fastfeedback without Meta object returned from this API.

Video clip available in [this comment link](https://intersective.atlassian.net/browse/AV2-750?focusedCommentId=45496)

If this fix is not necessary, please feel free to let me know too!

### DIFF
--- a/src/app/fast-feedback/fast-feedback-submitter.service.ts
+++ b/src/app/fast-feedback/fast-feedback-submitter.service.ts
@@ -12,8 +12,8 @@ export class FastFeedbackSubmitterService {
     private request: RequestService
   ) {}
 
-  submit(data, params) {
-    return this.request.post(api.submit, data, {params: params});
+  submit(data, params: {}) {
+    return this.request.post(api.submit, data, { params });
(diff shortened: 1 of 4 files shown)


Barry-Chen-Intersective:
@trtshen , https://intersective.atlassian.net/browse/AV2-753 is the story to fix it. Yes, we made changes to Backend pulse check list api, it has a side effect that if the slider config has alert as other values rather than `triggered`, it returns slider data. Front end needs to make some changes to that modal pop up logics and deal with the slider only if slider and meta both exist in the response data.

PR #981, opened 2020-09-22
Task/av2 697/get event list api use graphql

ronakamani:


src/app/event-list/event-list.component.ts
```diff
@@ -57,12 +58,13 @@ export class EventListComponent {
     this.selectedActivities = [];
     this.loadingEvents = true;
     this.activated = 'browse';
+    this.type = 'activity_session';
```
shawnm0705, line 61: Since the App doesn't have another type other than "activity_session", we don't need to set it up as a variable, just need to pass "activity_session" in the service level. Component shouldn't be aware of this

src/app/event-list/event-list.service.ts
```diff
@@ -70,23 +69,30 @@ export class EventListService {
...
-      params = {
-        type: 'activity_session'
-      };
-    }
-    return this.request.get(api.get.events, {params: params})
-      .pipe(map(response => {
-        return this.normaliseEvents(response.data);
-      })
-    );
+  getEvents(type, activity_Id?) {
```
shawnm0705, line 72: `activity_Id` -> `activityId`
`event_type` -> `eventType`

src/app/event-list/event-list.service.ts
```diff
@@ -96,67 +102,79 @@ export class EventListService {
...
-          url: event.video_conference.url,
-          meetingId: event.video_conference.meeting_id,
-          password: event.video_conference.password
-        } : null
+        remainingCapacity: event.remainingCapacity,
+        isBooked: event.isBooked,
+        singleBooking: event.singleBooking,
+        canBook: event.canBook,
+        isPast: this.utils.timeComparer(event.eventStart) < 0,
+        assessment: this.utils.has(event, 'assessment.id')
```
shawnm0705, line 140: In the query above, we just request for 
```
assessment {
  id name
}
```
how do we get context_id and is_done?

src/app/event-list/event-list.service.ts
```diff
@@ -96,67 +102,79 @@ export class EventListService {
...
+          ? {
+              id: event.assessment.id,
+              contextId: event.assessment.context_id,
+              isDone: event.assessment.is_done || false,
+            }
+          : null,
+        videoConference: this.utils.has(event, 'video_conference.url')
```
shawnm0705, line 147: check the doc for the key name

src/app/event-list/event-list.service.ts
```diff
@@ -96,67 +102,79 @@ export class EventListService {
...
 
   getSubmission(assessmentId, contextId): Observable<any> {
-    return this.request.get(api.get.submissions, {params: {
-        assessment_id: assessmentId,
-        context_id: contextId,
-        review: false
-      }})
-      .pipe(map(response => {
-        return !this.utils.isEmpty(response.data);
+    return this.request
```
shawnm0705, line 165: Are we still using this? Do we still need it?

src/app/tabs/tabs.component.ts
```diff
@@ -103,7 +103,7 @@ export class TabsComponent extends RouterEnter {
       this.showEvents = true;
     } else {
       this.showEvents = false;
-      this.eventsService.getEvents().subscribe(events => {
+      this.eventsService.getEvents('activity_session', null).subscribe(events => {
```
shawnm0705, line 106: Don't need to pass `null`

src/app/event-list/event-list.service.ts
shawnm0705, line 72: As mentioned above, since we don't have other types now, we should put default type inside this function so that we don't need to pass `activity_session` when we invoke this function

shawnm0705 (changes requested):

PR #1000, opened 2020-10-15
Feature/av2 766/update chat api calls to new graphql apis

sasangachathumal:
Jira story - https://intersective.atlassian.net/browse/AV2-766

### DIFF
--- a/src/app/chat/chat-room/chat-room.component.ts
+++ b/src/app/chat/chat-room/chat-room.component.ts
@@ -35,8 +35,6 @@ export class ChatRoomComponent extends RouterEnter {
   channelUuid: string;
   // message history list
   messageList: Message[] = [];
-  // chat member list
-  memberList: ChannelMembers[] = [];
   // the message that the current user is typing
   message: string;
   messagePageCursor = '';
@@ -91,7 +89,6 @@ export class ChatRoomComponent extends RouterEnter {
   onEnter() {
     this._initialise();
     this._subscribeToTypingEvent();
-    this._loadMembers();
     this._loadMessages();
     this._scrollToBottom();
   }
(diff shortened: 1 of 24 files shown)


Barry-Chen-Intersective (changes requested):
@sasangachathumal , should call with correct data failed

src/app/chat/chat-room/chat-room.component.ts
```diff
@@ -68,18 +71,12 @@ export class ChatRoomComponent extends RouterEnter {
...
-        this.utils.broadcastEvent('channel-id-update', {
-          previousId: this.channelId,
-          currentId: receivedMessage.channelId
-        });
+      if (receivedMessage.channelUuid !== this.channelUuid) {
+        return;
       }
       if (receivedMessage && receivedMessage.file) {
-        receivedMessage.preview = this.attachmentPreview(receivedMessage.file);
+        receivedMessage.fileObject = JSON.parse(receivedMessage.file);
```
shawnm0705, line 78: better to check if it is a string, if so, JSON.parse()

src/app/chat/chat.service.ts
```diff
@@ -88,39 +89,49 @@ export class ChatService {
...
       this.request.apiResponseFormatError('Chat format error');
       return [];
     }
-    if (data.length === 0) {
+    if (result.length === 0) {
       return [];
     }
     const chats = [];
-    data.forEach(chat => {
+    result.forEach(chat => {
```
shawnm0705, line 122: Since the attributes are exactly the same, we don't need to re-assign it. 
GraphQL will make sure we get what we asked for

src/app/chat/chat.service.ts
```diff
@@ -12,63 +12,64 @@ import { environment } from '@environments/environment';
  * @type {Object}
  */
 const api = {
-  getChatList: 'api/v2/message/chat/list.json',
-  getChatMessages: 'api/v2/message/chat/list_messages.json',
   createMessage: 'api/v2/message/chat/create_message',
   markAsSeen: 'api/v2/message/chat/edit_message'
```
shawnm0705, line 16: Let's remove all those since we are not using them anymore

shawnm0705 (changes requested):

shawnm0705 (approved):

Barry-Chen-Intersective (approved):

PR #1023, opened 2020-11-25
[AV2-793 Task/pn-unitest

trtshen:
https://intersective.atlassian.net/browse/AV2-793


trtshen:
hi @ronakamani please spend some time play around with test script i've done for PushNotificationService.

PR #1032, opened 2020-11-30
put sonar ci as a separated file

Barry-Chen-Intersective:


Barry-Chen-Intersective:
@Sunilsbcloud , I applied the docker image caches to save the costs from downloading the images at each run, although the process of loading images from cache takes around 1 minute. It also reduce the uncertain because sometimes pulling images from ECR might take longer than average.

Sunilsbcloud:
Looks good to me.

PR #1160, opened 2021-05-04
Av2 908/broken cards

ronakamani:
https://intersective.atlassian.net/browse/AV2-908

**Additional Comments:**
`Add here`

**Checklist:**

- [ x ] Unit test completed?: (Y/N)
- [ x ] Docs folder up to date?: (Y/N)
- [ x ] Add if anything missed: (Y/N)

All Checklist are okay.

### DIFF
--- a/src/app/shared/components/clickable-item/clickable-item.component.html
+++ b/src/app/shared/components/clickable-item/clickable-item.component.html
@@ -2,7 +2,7 @@
   'switcher-card-item': isSwitcherCard,
   'active': active
 }"  aria-label="clickable item" >
-  <ion-button class="accessible-btn" aria-label="switcher card" role="button">
+  <ion-button class="accessible-btn" [ngClass]="{'accessible-item': isCustomizedCard, 'accessible-chat': isChatCard}" aria-label="switcher card" role="button">
     <ion-text color="primary" [ngStyle]="{'color': backgroundColor}">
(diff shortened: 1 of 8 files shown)


src/app/shared/components/clickable-item/clickable-item.component.html
```diff
@@ -2,7 +2,7 @@
   'switcher-card-item': isSwitcherCard,
   'active': active
 }"  aria-label="clickable item" >
-  <ion-button class="accessible-btn" aria-label="switcher card" role="button">
+  <ion-button class="accessible-btn" [ngClass]="{'accessible-item': isCustomizedCard, 'accessible-chat': isChatCard}" aria-label="switcher card" role="button">
```
sasangachathumal: `accessible-chat` css class is an empty css class, it didn't have any styles in it. so why we need extra logic to check `isChatCard` and add that class.

sasangachathumal (changes requested):

src/app/shared/components/clickable-item/clickable-item.component.html
ronakamani: @sasangachathumal , it will be fixed in another story. :D sure I'll remove it here. thanks

sasangachathumal (approved):

PR #1172, opened 2021-05-19
add new settings route and make changes to assessment titlebar

jazzmind:
Created a page /settings-embed that can be used to show the settings without nav elements .

Changed assessment page so the titlebar reflects the state of the assessment better.

**Checklist:**

- [ x ] Unit test completed?: (Y/N)
- [ x ] Docs folder up to date?: (Y/N)
- [ x ] Add if anything missed: (Y/N)

All Checklist are okay.


src/app/auth/auth-direct-login/auth-direct-login.component.ts
```diff
@@ -85,6 +85,13 @@ export class AuthDirectLoginComponent implements OnInit {
       // switch to the program
       await this.switcherService.switchProgram(program).toPromise();
     }
+    const referrerUrl = this.route.snapshot.paramMap.get('activity_task_referrer_url');
```
shawnm0705: @jazzmind `activity_task_referrer_url` is meant to be the referrer URL for direct link to the activity_task page.
If we need the similar logic for assessment page, shall we name it `assessment_referrer_url` and put it under the assessment `case` accordingly?

src/app/settings-embed/settings-embed.component.ts
```diff
@@ -0,0 +1,151 @@
+import { Component } from '@angular/core';
```
shawnm0705, line 1: Generally, it's not a good idea to just copy & paste all the code and create a different route to it.
We should just pass a parameter to the settings component to hide/show the elements.

src/environments/environment.ts.wes
```diff
@@ -0,0 +1,51 @@
+// The file contents for the current environment will overwrite these during build.
```
shawnm0705: @jazzmind is this intended to be committed to git or should it be gitignored?

shawnm0705 (changes requested):

src/app/auth/auth-direct-login/auth-direct-login.component.ts
jazzmind: I think we should have a generic referrer URL for any direct link which is where the back arrow and similar behavior goes. It is not specific to the activity_task so it should just be referrer_url... however we should leave activity_task_referrer_url in there as an alternative for backward compatiblity.

src/app/settings-embed/settings-embed.component.ts
jazzmind: I know, but we need this quickly for xprize and altering the existing settings page was more work and would require more tests. I think we should do what you suggest as part of incorporating into the main app-v2 branch.

src/environments/environment.ts.wes
jazzmind: should be git ignored... oops

src/app/auth/auth-direct-login/auth-direct-login.component.ts
shawnm0705: The problem is how do we know which back button should be redirected to the referrer URL
e.g. we may have a use case that we want a user to navigate the activity page and assessment/topic page in app, only redirect user back to referrer when user click back button from activity page.
we may have another use case that we want a user to only see the assessment page in app, redirect user back to referrer when user click back button from assessment page.
We will need a way to handle this different scenario. That's why I put this activity_task referrer URL to the "activity task" specific case

jazzmind: Sounds like we need some design work here to list out use cases we want to support for embedding the app. I agree in the interim to call it assessment_referrer_url so that we preserve optionality.

jazzmind: I have made a change which i think gives us some flexibility and a level of abstraction... let me know what you think @shawnm0705

shawnm0705 (approved):

PR #1200, opened 2021-06-10
[AV2-922] Feature/deeplink-review

trtshen:
https://intersective.atlassian.net/browse/AV2-922

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.


sasangachathumal (dismissed):

src/app/auth/auth-direct-login/auth-direct-login.component.ts
```diff
@@ -59,27 +57,40 @@ export class AuthDirectLoginComponent implements OnInit {
...
     const assessmentId = +this.route.snapshot.paramMap.get('asmt');
     const submissionId = +this.route.snapshot.paramMap.get('sm');
     const topicId = +this.route.snapshot.paramMap.get('top');
+    let timelineId = +this.route.snapshot.paramMap.get('tl');
+
     // clear the cached data
     this.utils.clearCache();
+
+    // if timelineId "0", then try get cached timelineId
+    timelineId = (timelineId > 0) ? timelineId : this.storage.getUser().timelineId;
```
shawnm0705: @trtshen why do we need to get timeline id from storage?

src/app/tabs/tabs.component.ts
```diff
@@ -38,23 +39,26 @@ export class TabsComponent extends RouterEnter {
...
-    this.utils.getEvent('notification').subscribe(event => {
-      this.noOfTodoItems++;
-    });
-    this.utils.getEvent('event-reminder').subscribe(event => {
-      this.noOfTodoItems++;
-    });
-    this.utils.getEvent('chat:new-message').subscribe(event => {
-      this.tabsService.getNoOfChats().subscribe(noOfChats => {
-        this.noOfChats = noOfChats;
+    this.restrictedAccess = storage.get('singlePageAccess');
```
shawnm0705: storage -> this.storage ?

src/app/tabs/tabs.component.ts
```diff
@@ -38,23 +39,26 @@ export class TabsComponent extends RouterEnter {
...
+
+    if (this.restrictedAccess === false) {
```
shawnm0705, line 41: if we haven't set `singlePageAccess` in the storage, it will return `null` I think?

src/app/services/storage.service.ts
```diff
@@ -47,6 +47,12 @@ export interface Config {
   color?: string;
 }
 
+export interface General {
```
shawnm0705: We are not using this anywhere?

shawnm0705 (changes requested):

src/app/tabs/tabs.component.ts
trtshen: this is "constructor"

src/app/auth/auth-direct-login/auth-direct-login.component.ts
trtshen: the "tl" must not be optional?

src/app/tabs/tabs.component.ts
shawnm0705: I suggest we create a function in storage.service `getRestrictedAccess()` like for other keys, so that we make sure it returns false if not set

src/app/auth/auth-direct-login/auth-direct-login.component.ts
trtshen: @shawnm0705 the "tl" cannot be optional?

src/app/tabs/tabs.component.ts
trtshen: yes, i thought of that too, but i just feeling something isn't right, because we have too much similar purpose purpose function with different name.
- getUser
- setUser
- getReferrer
- setReferrer
- getConfig
- setConfig
- getBooked....ids
- setBooked....ids
- getCountry
- setCountry
- setCurrentChatChannel
- getCurrentChatChannel

and all of them are using the "get" & "set" functions in the almost same way.

trtshen: i can follow the same pattern, but it actually worsen the maintainability & it contributes to the tech debt?

src/app/services/storage.service.ts
trtshen: i added this, to build up the structure that we are using for the entire localStorage. Right now, there is no direct way to see what is being cache from the high level view.
you created interface for "User", which is very good way to trace what is being cached and what isn't. 

Other than that, we lost track of the "general" view of the entire cache, so even experienced developer may find it hard to really sure what can be used or add in the highest level of the cache.

shawnm0705 (approved):

PR #1206, opened 2021-06-28
[AV2-928] Feature/get stack info from stack UUID

trtshen:
https://intersective.atlassian.net/browse/AV2-928

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -138,9 +138,6 @@ export class AppComponent implements OnInit {
     }
 
     const stackUuid = searchParams.get('stack_uuid');
-    console.log('stackUuid:', stackUuid);
-
-
     if (stackUuid) {
       this.retrieveStackConfig(stackUuid);
     }
@@ -178,14 +175,11 @@ export class AppComponent implements OnInit {
   retrieveStackConfig(stackUuid: string): void {
     let stack: any;
     if (stackUuid) {
-  console.log('asdasdasd::', stackUuid);
       this.authService.getStackConfig(stackUuid).subscribe(res => {
-        console.log(res);
-
         this.storage.stackConfig = res;
       });
     }
 
-    return stack;
+    return this.storage.stackConfig;
   }
 }
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -173,7 +173,6 @@ export class AppComponent implements OnInit {
    * @return  {void}
    */
   retrieveStackConfig(stackUuid: string): void {
-    let stack: any;
     if (stackUuid) {
       this.authService.getStackConfig(stackUuid).subscribe(res => {
         this.storage.stackConfig = res;
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -30,8 +30,6 @@ export class AppComponent implements OnInit {
     private ngZone: NgZone,
     private newRelic: NewRelicService,
     public sanitizer: DomSanitizer,
-    // private splashScreen: SplashScreen,
-    // private statusBar: StatusBar
   ) {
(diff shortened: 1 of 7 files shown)


src/app/app.component.ts
```diff
@@ -138,6 +136,11 @@ export class AppComponent implements OnInit {
           break;
       }
     }
+
+    const stackUuid = searchParams.get('stack_uuid');
+    if (stackUuid) {
+      this.retrieveStackConfig(stackUuid);
```
shawnm0705, line 142: we need to put `await` on this, otherwise we won't wait for it

src/app/app.component.ts
```diff
@@ -162,4 +165,20 @@ export class AppComponent implements OnInit {
...
+   * @return  {void}
+   */
+  retrieveStackConfig(stackUuid: string): void {
+    if (stackUuid) {
+      this.authService.getStackConfig(stackUuid).subscribe(res => {
+        this.storage.stackConfig = res;
+      });
+    }
+
+    return this.storage.stackConfig;
```
shawnm0705, line 182: we don't need to return it here. If we just return it here, it doesn't have the value from the API

src/app/auth/auth.service.ts
```diff
@@ -22,7 +22,11 @@ const api = {
   register: 'api/registration_details.json',
   forgotPassword: 'api/auths.json?action=forgot_password',
   verifyResetPassword: 'api/auths.json?action=verify_reset_password',
-  resetPassword: 'api/auths.json?action=reset_password'
+  resetPassword: 'api/auths.json?action=reset_password',
+};
+
+const LOGIN_API = {
+  stackInfo: 'https://login.practera.com/stack',
```
shawnm0705, line 29: the domain should be stored in the environment.ts instead of hardcode here

src/app/auth/auth.service.ts
```diff
@@ -331,10 +359,26 @@ export class AuthService {
...
+   * get stack information by uuid through LoginAPI
+   *
+   * @param   {string<StackConfig>}      uuid
+   *
+   * @return  {Observable<StackConfig>}        observable response of stack endpont
+   */
+  getStackConfig(uuid: string): Observable<StackConfig> {
+    return this.request.get(LOGIN_API.stackInfo, { uuid }).pipe(map(res => {
+      if (res && res.data) {
+        return res.data;
```
shawnm0705, line 379: the response doesn't have `data` field, it is the stack object itself

src/app/services/storage.service.ts
```diff
@@ -167,4 +167,13 @@ export class BrowserStorageService {
   set singlePageAccess(val) {
     this.set('singlePageAccess', val);
   }
+
+  get stackConfig() {
+    const result = this.get('stackConfig');
+    return result || false;
```
shawnm0705, line 173: should not be `false` by default, it should be an object, so default is `null` or `{}`, and that is covered in the `this.get()` already, so don't need to specify it here.

src/app/auth/auth.service.ts
```diff
@@ -331,10 +359,26 @@ export class AuthService {
...
+
+  /**
+   * get stack information by uuid through LoginAPI
+   *
+   * @param   {string<StackConfig>}      uuid
+   *
+   * @return  {Observable<StackConfig>}        observable response of stack endpont
+   */
+  getStackConfig(uuid: string): Observable<StackConfig> {
+    return this.request.get(LOGIN_API.stackInfo, { uuid }).pipe(map(res => {
```
shawnm0705, line 377: the `uuid` is in the wrong place, it should be put into `params`

PR #1207, opened 2021-06-28
[AV2-924] Update Login API Call

sasangachathumal:
added urlFormatter helper.
added urlformatter to request.service.ts
Update auth service login functions.
Update login component email to username.
Update registration component to work with new login flow.
updated request interceptor not to pass appkey to login api.
updated environment files.

https://intersective.atlassian.net/browse/AV2-924

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.


src/app/auth/auth.service.ts
```diff
@@ -77,6 +88,11 @@ export class AuthService {
     // do clear user cache here
   }
 
+  private _loginAPILogin(body: LoginRequParams) {
+    body.from = 'App';
+    return this.request.loginAPIPost(api.loginAPI.login, body);
```
trtshen: @sasangachathumal  why is the different between writing this as 
`return this.request.post(api.loginAPI.login + 'login', body);` 

compared to this code in line 93?

Could you please explain why is it necessary to clone the [entire post function](https://github.com/intersective/app-v2/blob/AV2-924/update-login-api-call/src/app/shared/request/request.service.ts#L142-L165)?

sasangachathumal: we still need both logins. when we login using login API, it return apikey right? but we can’t use it to call core APIs. we need to call core API login with that apikey to get apikey for login. that’s what i ask from shown in standup yesterday. that why i can’t just simply replace our core login with login api. i need to add new method to call login API.

it same like if user coming to appv2 from global login, url contain api key, in global login method we call core api login with that apikey to get correct one.

login API return an global apikey, but to call stack apis we need to get api key from that stack core api.

sasangachathumal: and also it, it's calling login API `login` endpoint.
`return this.request.post(api.loginAPI.login, body);`
not this
`return this.request.post(api.loginAPI.login + 'login', body);`

sasangachathumal: you can see api const values.
https://github.com/intersective/app-v2/pull/1207/files?file-filters%5B%5D=.ts#diff-0cc1fef6a75cb019a3dd68c49beec2725b302d53b51a98321f908b5da6cbe518R25-R28

trtshen (changes requested):
https://github.com/intersective/app-v2/pull/1207/files#r659887207

src/app/auth/auth-registration/auth-registration.component.ts
```diff
@@ -182,15 +182,26 @@ export class AuthRegistrationComponent implements OnInit {
...
-                  this.showPopupMessages('shortMessage', 'Registration success!', route);
+                async globalRes => {
+                  this.authService.globalLogin({
+                    apikey: globalRes.apikey,
+                    service: 'LOGIN'
+                  }).subscribe(
+                  async res => {
+                    nrAutoLoginTracer();
+                    this.storage.remove('unRegisteredDirectLink');
+                    const route = await this.switcherService.switchProgramAndNavigate(res.programs);
```
shawnm0705: Global login will not return the programs, we need to get the programs from core POST auth

shawnm0705 (changes requested):
@sasangachathumal we'd better put some comments of the workflow for each auth related endpoint. e.g. for the login workflow, what's first, what's next.
Either put the comments as a docblock of the function, or put then in a stand alone .md docs file

src/app/auth/auth-registration/auth-registration.component.ts
sasangachathumal: method call `globalLogin` not sure why. it created before. but it calling core API /auth.json.

src/helper.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Given a domain and endpoint string, return a correctly formatted url string
+ * e.g.
+ * test.practera.com - login => https://test.practera.com/login
+ *
+ */
+export const urlFormatter = (domain: string, endpoint?: string) => {
```
trtshen: @sasangachathumal i understand we use `helper.ts` file in our loosely structured project, our angular app has obvious file structure to follow. This utility-related function should placed inside UtilsService file.

trtshen (changes requested):

src/helper.ts
sasangachathumal: sure

src/app/auth/auth-registration/auth-registration.component.ts
sasangachathumal: @shawnm0705 I added comment in code and also i created a .md file in './docs/workflows/auth-workflows.md'. and mention in README.md file and also mention in code comments.

and i rename the `globalLogin` to `directLoginWithApikey`.

src/helper.ts
sasangachathumal: @trtshen I have moved 'urlFormatter' to the utilService.

trtshen (approved):

src/app/auth/auth.service.ts
```diff
@@ -104,7 +111,27 @@ export class AuthService {
...
+  /**
+   * @name _loginAPILogin
+   * @description Calling login API to login the user and save stacks in local storage.
+   * @param body Json Object - request parameter need to pass to login api.
+   */
+  private _loginAPILogin(body: LoginRequParams): Observable<any> {
+    body.from = 'App';
+    return this.request.post(LOGIN_API.login, body, {}, true)
+    .pipe(map(res => {
+      this.storage.set('stacks', res.stacks);
```
shawnm0705: the logic to save the stacks in local storage should be inside the component instead of the service.
Also, let's not use `this.storage.set()` directly, as it makes it hard to track what has been stored in the local storage, let's create a get and set function in storage.service same as others for this.

We'd better also specify the type of the stack as well, so that we know what attributes does the stack has.
Chaw has `StackConfig` in the auth.service, please talk to him to move it to the storage.service instead, and let's just rename it to `Stack`

src/app/auth/auth-login/auth-login.component.ts
```diff
@@ -60,54 +69,57 @@ export class AuthLoginComponent implements OnInit {
...
-      email: this.loginForm.value.email,
+      username: this.loginForm.value.username,
       password: this.loginForm.value.password,
     }).subscribe(
-      res => {
-        nrLoginTracer('login successful');
-        this.newRelic.actionText('login successful');
-        return this._handleNavigation(res.programs);
+      globalRes => {
+        return this.authService.directLoginWithApikey({
```
shawnm0705: the logic here is still incorrect. After we get the stacks, we store them in local storage and redirect user to experience switcher page. Then we need to make `auth` requests for each stacks to get the experiences and gather them together.

shawnm0705 (changes requested):

PR #1215, opened 2021-07-05
[AV2-925] Update Forget Password API Call

sasangachathumal:
added new stack objects in unit test.
added forgot password to auth service.

https://intersective.atlassian.net/browse/AV2-925

**Additional Comments:**
`Add here`

**Checklist:**

- [x] Unit test completed?: (Y)
- [x] Docs folder up to date?: (Y)

All checklists are okay.

### DIFF
--- a/src/app/auth/auth.service.ts
+++ b/src/app/auth/auth.service.ts
@@ -19,14 +19,14 @@ const api = {
   setProfile: 'api/v2/user/enrolment/edit.json',
   verifyRegistration: 'api/verification_codes.json',
   register: 'api/registration_details.json',
-  forgotPassword: 'api/auths.json?action=forgot_password',
   verifyResetPassword: 'api/auths.json?action=verify_reset_password',
   resetPassword: 'api/auths.json?action=reset_password',
 };
 
 const LOGIN_API = {
   stackInfo: 'stack',
-  login: 'login'
+  login: 'login',
+  forgotPassword: 'forgotPassword'
 };
 
 interface VerifyParams {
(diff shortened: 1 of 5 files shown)


src/app/auth/auth.service.ts
```diff
@@ -227,10 +227,23 @@ export class AuthService {
...
    */
   forgotPassword(email: string): Observable<any>  {
-    return this.request.post(api.forgotPassword, {
-      email: email,
-      domain: this.getDomain()
-    });
+    const body = {
+      email,
+      resetLink: this._createResetDirectLinks(),
+      directLink: this._createResetDirectLinks('direct')
```
shawnm0705: why not just separate them into two different functions.
Or make the function name more generic like `_createLinks`

shawnm0705 (changes requested):

src/app/shared/request/request.service.ts
```diff
@@ -96,7 +96,7 @@ export class RequestService {
   }
 
   private getEndpointUrl(endpoint, isLoginAPI?: boolean) {
-    let endpointUrl = this.prefixUrl + endpoint;
+    let endpointUrl = ( this.storage.stackConfig ? this.storage.stackConfig.coreApi : this.prefixUrl ) + endpoint;
```
shawnm0705: there is no `this.prefixUrl` anymore, it should be removed from here an the app module as well.
If there is no stack url, we should throw error.

shawnm0705 (changes requested):

shawnm0705 (approved):

PR #1216, opened 2021-07-08
Feature/av2 926/update reset password api call

sasangachathumal:
https://intersective.atlassian.net/browse/AV2-926

**Additional Comments:**
`Add here`

**Checklist:**

- [x] Unit test completed?: (Y)

All checklists are okay.

### DIFF
--- a/src/app/auth/auth.service.ts
+++ b/src/app/auth/auth.service.ts
@@ -392,8 +392,8 @@ export class AuthService {
    */
   getStackConfig(uuid: string): Observable<Stack> {
     return this.request.get(LOGIN_API.stackInfo, {params: {uuid}}, true).pipe(map(res => {
-      if (res && res.data) {
-        return res.data;
+      if (res) {
+        return res;
       }
       return null;
     }));
--- a/src/app/auth/auth.service.ts
+++ b/src/app/auth/auth.service.ts
@@ -19,7 +19,6 @@ const api = {
   setProfile: 'api/v2/user/enrolment/edit.json',
   verifyRegistration: 'api/verification_codes.json',
   register: 'api/registration_details.json',
-  verifyResetPassword: 'api/auths.json?action=verify_reset_password',
 };
 
 const LOGIN_API = {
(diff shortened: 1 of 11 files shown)


src/app/auth/auth.service.ts
```diff
@@ -20,13 +20,13 @@ const api = {
   verifyRegistration: 'api/verification_codes.json',
   register: 'api/registration_details.json',
   verifyResetPassword: 'api/auths.json?action=verify_reset_password',
```
shawnm0705: If we don't need this anymore, please delete it

trtshen (approved):

shawnm0705 (changes requested):

src/app/shared/request/request.service.ts
```diff
@@ -98,16 +98,16 @@ export class RequestService {
...
     if (isLoginAPI) {
-      endpointUrl = this.loginApiUrl + endpoint;
+      endpointUrl = this.utils.urlFormatter(this.loginApiUrl, endpoint);
     } else if (this.storage.stackConfig && this.storage.stackConfig.coreApi) {
-      endpointUrl = this.storage.stackConfig.coreApi + endpoint;
+      endpointUrl = this.utils.urlFormatter(this.storage.stackConfig.coreApi, endpoint);
     } else {
       throw new Error('Can not find API URL.');
     }
     if (endpoint.includes('https://') || endpoint.includes('http://')) {
```
shawnm0705: since we have the urlFormatter, we don't need to check the http here anymore

shawnm0705 (changes requested):

src/app/shared/request/request.service.ts
sasangachathumal: @shawnm0705 
in some places we pass full url as endpoint ( "http://abcd.com/123" ). like the URL we use to get user location ("https://ipapi.co/json"). this is for check that. to not add our API URL in to those urls.
if endpoint have http or https then not adding any url to endpoint.

shawnm0705: inside the urlFormatter, we also checks if it has `http`, so this is redundant I think?

shawnm0705 (changes requested):

src/app/shared/request/request.service.ts
sasangachathumal: I can't find any place we use this like this. let's remove this as you mention, if there any issue came we can fix this later

shawnm0705 (approved):

PR #1217, opened 2021-07-11
[AV-842] Feature/localstorage stacks

trtshen:
https://intersective.atlassian.net/browse/AV2-842

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.


src/app/auth/auth-login/auth-login.component.ts
```diff
@@ -68,7 +55,16 @@ export class AuthLoginComponent implements OnInit {
...
-      res => {
+      (res: {
+        apikey: string;
+        stacks: Stack[];
+      }) => {
+        if (res.stacks && res.stacks.length === 0) {
+          return this.notificationFormat('No stacks available for this user.');
+        }
+
+        // this.storage.stackConfig = res.stacks[0] ? res.stacks[0] : null;
```
sasangachathumal: can we remove this commented code line. if it no need in the future also.

src/app/auth/auth.service.ts
```diff
@@ -94,6 +94,7 @@ export class AuthService {
   private _loginFromLoginAPI(body: LoginRequParams): Observable<any> {
     body.from = 'App';
     return this.request.post(LOGIN_API.login, body, {}, true);
+    // return this.request.post(LOGIN_API.login, body, {}, true);
```
sasangachathumal: can we remove this commented code line. if it no need.

shawnm0705: duplicate line

sasangachathumal (changes requested):

src/app/auth/auth-login/auth-login.component.ts
```diff
@@ -68,7 +55,16 @@ export class AuthLoginComponent implements OnInit {
       username: this.loginForm.value.username,
       password: this.loginForm.value.password,
     }).subscribe(
-      res => {
+      (res: {
+        apikey: string;
+        stacks: Stack[];
+      }) => {
+        if (res.stacks && res.stacks.length === 0) {
+          return this.notificationFormat('No stacks available for this user.');
```
shawnm0705: User is not aware of "stack", so we shouldn't display message including the "stack"

src/app/auth/auth-login/auth-login.component.ts
shawnm0705: unused code

shawnm0705 (changes requested):

sasangachathumal (approved):

src/app/switcher/switcher-program/switcher-program.component.ts
```diff
@@ -23,15 +24,23 @@ import { UtilsService } from '@services/utils.service';
...
     private switcherService: SwitcherService,
     private newRelic: NewRelicService,
     private notificationService: NotificationService,
-    private utils: UtilsService
-  ) { super(router); }
+    private utils: UtilsService,
+    private activatedRoute: ActivatedRoute,
+  ) {
+    super(router);
+    activatedRoute.data.subscribe(data => {
```
shawnm0705: We don't need to get the stacks from the resolver, we will get the stacks from the local storage anyway.

shawnm0705 (changes requested):

src/app/auth/auth.service.ts
```diff
@@ -384,4 +387,31 @@ export class AuthService {
...
+        if (res) {
+          this.storage.stacks = res;
+        }
+        return res;
+      }),
+      map(res => {
+        if (res) {
+          return res;
+        }
+        return [];
```
shawnm0705: can we combine those into one pipe?

shawnm0705 (changes requested):

src/app/switcher/switcher-program/switcher-program.component.ts
trtshen: hi @shawnm0705 thanks for your suggestion, but both resolver and localstorage serve the same purpose.
Resolver has better advantages where the localStorage fallshort. Resolver makes sure "stacks" get pulled correctly before user get redirected to the page, so switcher page is always guaranteed with "stacks" value to started with.

Unless the speed of one single API call to `https://login.practera.com/stacks` concerns you, otherwise we shouldn't be too worried about the loading speed, anything other than this `https://login.practera.com/stacks` API call is still happen inside our switcher loading screen.

src/app/auth/auth.service.ts
trtshen: sure, can do

PR #1261, opened 2021-09-01
[AV2-950] Bugfix/chat-display-when-channel-exist

trtshen:
https://intersective.atlassian.net/browse/AV2-950

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.

### DIFF
--- a/src/app/overview/home/home.component.ts
+++ b/src/app/overview/home/home.component.ts
@@ -11,7 +11,6 @@ import { NewRelicService } from '@shared/new-relic/new-relic.service';
 import { trigger, state, transition, style, animate, useAnimation } from '@angular/animations';
 import { fadeIn } from '../../animations';
 import { Observable, Subscription } from 'rxjs';
-import { ChatService } from '@app/chat/chat.service';
 
 @Component({
   selector: 'app-home',
@@ -53,7 +52,6 @@ export class HomeComponent implements OnDestroy, OnInit {
     public achievementService: AchievementsService,
     private eventsService: EventListService,
     private newRelic: NewRelicService,
-    readonly chatService: ChatService,
   ) {
     const role = this.storage.getUser().role;
     this.utils.getEvent('notification').subscribe(event => {
--- a/src/app/overview/home/home.component.ts
+++ b/src/app/overview/home/home.component.ts
@@ -51,7 +51,7 @@ export class HomeComponent implements OnDestroy, OnInit {
     public storage: BrowserStorageService,
     public achievementService: AchievementsService,
     private eventsService: EventListService,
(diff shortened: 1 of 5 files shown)


src/app/overview/home/home.component.ts
```diff
@@ -108,17 +110,20 @@ export class HomeComponent implements OnDestroy, OnInit {
...
-        this.homeService.getChatMessage().subscribe(chatMessage => {
-          if (!this.utils.isEmpty(chatMessage)) {
-            this._addChatTodoItem(chatMessage);
-          }
-          this.loadingTodoItems = false;
-        })
-      );
-    }
+    // only get the number of chats if user has chatroom
+    this.chatService.getChatList().subscribe(chats => {
```
shawnm0705: we don't need to check this here. Previously we check if user is in team because if user is not in team, we can save the time to send an API request, since "check if user is in team" doesn't need an API request.
Now it's unnecessary to make an additional API request to check if user has channels. We can just send the request to get chat messages

src/app/tabs/tabs.component.ts
```diff
@@ -63,9 +65,9 @@ export class TabsComponent extends RouterEnter {
   }
 
   private _initialise() {
-    this.showChat = false;
-    this.showReview = false;
-    this.showEvents = false;
+    this.showChat = this.showChat || false;
+    this.showReview = this.showReview || false;
+    this.showEvents = this.showEvents || false;
```
shawnm0705: If we change it to this, we may have the following problem:
1. user 1 login and should see chat/review/events, he will see them
2. user 1 logout
3. user 2 login and shouldn't see chat/review/events, but he will see them at first because those 3 variables are true

shawnm0705 (changes requested):

src/app/tabs/tabs.component.ts
trtshen: oh! i noticed there is a ugly UX problem:
switching between pages will cause tabs keep getting hide and show.

the chance that a 2nd user login and see the tab getting hidden once is still a better solution than keep showing the "UX problem" to every user. What do you think @shawnm0705 ?

shawnm0705: @trtshen in that case, we need to store a variable in local storage to indicate if we should display those in tab. And remember to clear it when user logout

trtshen: @shawnm0705 no need, i've tested this 3 lines changes, works well enough for fixing "jumpy tabs", but i would like to know what's your plan why do we need localStorage to handle this? However, these values should be safely "refreshed" (disposable) at every page refresh.

PR #1309, opened 2021-10-22
[CORE-4930, CORE-4959] Full events changes need to do in App

sasangachathumal:
https://intersective.atlassian.net/browse/CORE-4930
https://intersective.atlassian.net/browse/CORE-4959

**Additional Comments:**
### **Check Jura story videos to check how changes works in app.**

Update event book button to show fully booked if no capacity remaining
Added code to disabled fully booked if no capacity remaining.
updated unit test for new changes.
update code to reduce capacity by 1 when event got booked.
Add code to show fully booked in events list if no capacity remaining.

**Checklist:**

All checklists are okay.

### DIFF
--- a/src/app/event-detail/event-detail.component.ts
+++ b/src/app/event-detail/event-detail.component.ts
@@ -130,6 +130,9 @@ export class EventDetailComponent implements OnInit {
         if (this.event.singleBooking) {
           this.storage.setBookedEventActivityIds(this.event.activityId);
         }
+        // update event remaining capacity
+        this.event.remainingCapacity -= 1;
+
         this.ctaIsActing = false;
       },
       error => {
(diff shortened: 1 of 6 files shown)


src/app/event-detail/event-detail.component.ts
```diff
@@ -130,6 +130,9 @@ export class EventDetailComponent implements OnInit {
         if (this.event.singleBooking) {
           this.storage.setBookedEventActivityIds(this.event.activityId);
         }
+        // update event remaining capacity
+        this.event.remainingCapacity -= 1;
```
shawnm0705: why do we need to reduce the remainingCapacity here?

shawnm0705 (changes requested):

src/app/event-detail/event-detail.component.ts
sasangachathumal: Main reason is, it' mention in the requirements of the story CORE-4930 and also in main story CORE-4853.
`When a booking is done by a user (learner or expert) on AppV2, AppV2’s max capacity shall reduce 1 unit`.

But I just check in stage environment it's reduce remaining capacity. So i think we can remove this code line. i will update this shawn.

shawnm0705: That statement is what a user should see as the result. 
The reduction should happen in the API instead of the front-end.

shawnm0705 (approved):

PR #1315, opened 2021-11-01
core-4942-allowed-booking-without-capacity

mehsan1:
allowed event booking without capacity


sasangachathumal:
My Changes.

- update code to allow event booking when capacity is null.
- capacity null means capacity in unlimited according to new events logic.

**Note**: API still returning error so need to wait until API fixed this to merge trunk

PR #1317, opened 2021-11-11
[AV2-979] Review Rating Scrollbar not helpful in its current format.

sasangachathumal:
`Add link here`

**Additional Comments:**
Update ion-range component to show notches in slider.
create method to round up values slider give before send to API.

**Checklist:**

All checklists are okay.

### DIFF
--- a/src/app/review-rating/review-rating.component.html
+++ b/src/app/review-rating/review-rating.component.html
@@ -10,7 +10,7 @@
   <h2 class="title">How helpful was the feedback?</h2>
 
   <div>
-    <ion-range mode="ios" min="0" max="1" step="0.1" snaps="true" [(ngModel)]="ratingData.rating" (ngModelChange)="roundUpRating()"></ion-range>
+    <ion-range mode="ios" min="0" max="1" step="0.1" snaps="true" [(ngModel)]="ratingData.rating"></ion-range>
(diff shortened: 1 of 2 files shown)


src/app/review-rating/review-rating.component.html
```diff
@@ -10,7 +10,7 @@ <h1 class="title">Rating</h1>
   <h2 class="title">How helpful was the feedback?</h2>
 
   <div>
-    <ion-range min="0" max="1" step="any" [(ngModel)]="ratingData.rating"></ion-range>
+    <ion-range mode="ios" min="0" max="1" step="0.1" snaps="true" [(ngModel)]="ratingData.rating" (ngModelChange)="roundUpRating()"></ion-range>
```
shawnm0705: we can just round up the rating when we post the data, don't need to do the round up every time when the value changes

shawnm0705 (changes requested):

src/app/review-rating/review-rating.component.html
sasangachathumal: yes. you are correct. and we already have that. what i add is a extra thing. it can removed from the code. I will update the code.

shawnm0705 (approved):

PR #1319, opened 2021-11-15
[AV2-983] hotfix/navigation-fix

trtshen:
https://intersective.atlassian.net/browse/AV2-983

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.


src/app/assessment/assessment.component.ts
```diff
@@ -257,18 +257,25 @@ export class AssessmentComponent extends RouterEnter {
                 {
                   text: 'OK',
                   role: 'cancel',
-                  handler: () => {
+                  handler: async () => {
                     if (this.activityId) {
-                      this._navigate(['app', 'activity', this.activityId ]);
+                      let direction = ['app', 'activity', this.activityId];
+
+                      // Due to onSameUrlNavigation: "ignore", unintended same URL navigation is now default to "/app/home".
```
shawnm0705, line 264: I think this logic is incorrect.
On desktop, the user will be on the activity page (i.e. `/app/activity/{activityId}`) . If there are other tasks inside this activity, we shouldn't redirect the user to the home page, we should let the user stay on this page.

So this logic only applies to when this team assessment is the only task inside this activity.

shawnm0705 (changes requested):

PR #1515, opened 2022-07-05
[AV2-1196] (V3, APP)Disable submit button when auto save is in progress.

sasangachathumal:
https://intersective.atlassian.net/browse/AV2-1196

**Additional Comments:**
update the auto save time out.
disable button when submission is lock.
disable button when saveInProgress true.

**Checklist:**

All checklists are okay.

### DIFF
--- a/projects/v3/src/app/components/assessment/assessment.component.ts
+++ b/projects/v3/src/app/components/assessment/assessment.component.ts
@@ -7,7 +7,7 @@ import { BrowserStorageService } from '@v3/services/storage.service';
 import { SharedService } from '@v3/services/shared.service';
 import { DOCUMENT, ViewportScroller } from '@angular/common';
 
-const SAVE_PROGRESS_TIMEOUT = 5000;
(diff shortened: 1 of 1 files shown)


projects/v3/src/app/components/assessment/assessment.component.ts
```diff
@@ -267,6 +270,7 @@ export class AssessmentComponent implements OnChanges {
 
     if (saveInProgress) {
       this.savingMessage = 'Saving...';
+      this.btnDisabled = true;
```
shawnm0705: We need to disable the button not only when save in progress, but also when submit assessment

shawnm0705 (changes requested):

projects/v3/src/app/components/assessment/assessment.component.ts
sasangachathumal: @shawnm0705 it's already doing by this method `continueToNextTask()`. this is the method call from HTML Button.

shawnm0705 (approved):

PR #1609, opened 2022-10-26
[AV2-1269]Event should be visible only to Participants

sasangachathumal:
https://intersective.atlassian.net/browse/AV2-1269

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.

### DIFF
--- a/projects/v3/src/app/pages/v3/v3.page.ts
+++ b/projects/v3/src/app/pages/v3/v3.page.ts
@@ -62,6 +62,8 @@ export class V3Page implements OnInit, OnDestroy {
   subscriptions: Subscription[];
   appPages: any[];
   showMessages: boolean = false;
+  showEvents: boolean = false;
+  showReviews: boolean = false;
   directionIcon: string = this.direction();
 
   constructor(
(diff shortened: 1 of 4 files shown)


shawnm0705 (approved):

projects/v3/src/app/pages/v3/v3.page.ts
```diff
@@ -105,40 +106,28 @@ export class V3Page implements OnInit, OnDestroy {
...
-          this._menuWithoutReview();
+          this.showReviews = false;
         }
       })
     );
+
     this.subscriptions.push(this.route.params.subscribe(_params => {
       this.reviewService.getReviews();
+
+      // Hide events tab to other user roles. Show only for participants
```
trtshen, line 125: hi @sasangachathumal i prefer this logic get written outside of `this.route.params` subscription.
Writing inside `this.route.params` may give us the false sense of the `role` is a dynamic value while it isn't in our appv3, because when a user login, their role would never change until a different user login to our app.

sasangachathumal: @trtshen I try to put it outside `this.route.params` but when it outside the `this.route.params`, check for the role not get executed when every time page load. But i will check again what i can do to improve this. Main reason i put it there `this.route.params` get executed every time page load or refreshed.

PR #1616, opened 2022-11-08
[AV2-1280] Hotfix/retain-appv2-review

trtshen:


trtshen:
no longer supporting v2

PR #1694, opened 2023-02-14
[AV2-1343] Bugfix/double uploads

trtshen:


trtshen:
reason of skip approval: involved changes are only translation script update

shawnm0705 (approved):

PR #1696, opened 2023-02-19
[AV2-1351] Task - locale detection & redirection

trtshen:


projects/v3/src/app/services/experience.service.ts
```diff
@@ -387,4 +387,26 @@ export class ExperienceService {
     return this.requestService.get(api.get.jwt);
   }
 
+  getLocale(): Observable<any> {
+    const currentURL = this.utils.getCurrentLocation();
+    if (currentURL.origin.includes('localhost')) { // ignore development
```
shawnm0705: shouldn't that be controlled over the environment variable instead?
e.g. use `environments.production`

projects/v3/src/app/services/experience.service.ts
```diff
@@ -387,4 +387,26 @@ export class ExperienceService {
...
+      return of({
+        locale: 'en-US'
+      });
+    }
+
+    return this.apolloService.graphQLFetch(
```
shawnm0705: we already have this query in home.service. We'd better reuse that one

shawnm0705 (changes requested):

shawnm0705 (approved):

PR #1789, opened 2023-05-09
[AV2-1259] Task/asmt mobile page unit test

trtshen:
Note: this unit-test is added so we can compare with new assessment api integration later.


trtshen:
I'm merging directly, as it only involves only unit test changes

PR #1793, opened 2023-05-16
[AV2-1376] add hubspot support form and questions popup

sasangachathumal:
https://intersective.atlassian.net/browse/AV2-1376

**Additional Comments:**
`Add here`

**Checklist:**

All checklists are okay.

### DIFF
--- a/projects/v3/src/app/services/experience.service.ts
+++ b/projects/v3/src/app/services/experience.service.ts
@@ -248,7 +248,6 @@ export class ExperienceService {
 
     // initialise Pusher
     this.sharedService.initWebServices();
-    this.utils.checkIsPracteraSupportEmail();
     try {
       const jwt = await this.getNewJwt().toPromise();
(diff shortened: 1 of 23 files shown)


projects/v3/src/app/services/experience.service.ts
```diff
@@ -248,6 +248,7 @@ export class ExperienceService {
 
     // initialise Pusher
     this.sharedService.initWebServices();
+    this.utils.checkIsPracteraSupportEmail();
```
trtshen: @sasangachathumal this type of initiator, it's recommended to move it to the `initWebServices()` as the name suggests, it's for initiations which is also suitable for the email support checking too.

trtshen (approved):

projects/v3/src/app/services/experience.service.ts
sasangachathumal: @trtshen Yes it's need to move. Will do that change.

PR #2031, opened 2023-11-08
[CORE-5977] Feature/new-authendpoint

trtshen:


trtshen:
@shawnm0705 i'll let you review and manually merge this PR, because you have control and know when the API get released at backend level on prerelease stack.

PR #2067, opened 2024-01-04
[CORE-6308] page title update on visit - prerelease (release 2.3.1)

trtshen:


projects/v3/src/app/services/activity.service.ts
```diff
@@ -294,6 +294,8 @@ export class ActivityService {
     if (!getData) {
       return ;
     }
+
+    this.utils.setPageTitle(task.name);
```
shawnm0705, line 298: do you mean to set up the page title inside a `service` file?

shawnm0705 (changes requested):
I think we will need the title for each page. Please add a proper title for each route

projects/v3/src/app/services/activity.service.ts
trtshen: yes, if you check, this is the source of truth of every topic/assessment page (mobile and desktop) are using to access/switch task. You can trace back to see where it's being used in the component, this is done for DRY practice.

shawnm0705 (approved):

PR #2212, opened 2024-08-01
[CORE-6721] bugfix/registration-wrong-redirection

trtshen:
https://intersective.atlassian.net/browse/CORE-6721


trtshen:
Note: merge directly so QA can start testing now.

PR #2218, opened 2024-08-08
[CORE-6673] allow redundant markAsDone

trtshen:
https://intersective.atlassian.net/browse/CORE-6673


trtshen:
note: close this one for better branch naming with semantic version

PR #2245, opened 2024-09-10
some traffic light code fixes

trtshen:


trtshen:
@jazzmind no code breaking now, so you can proceed to next step of the pulsecheck api implementation.
Screenshot 2024-09-10 at 2 32 40 PM

trtshen:
Added a little more changes to display the `on track` label, still not fully working as API query cannot return the confidence value yet.
Screenshot 2024-09-18 at 2 27 02 pm

PR #2271, opened 2024-10-15
[CORE-7535] achievemnt with graphql - prerelease

trtshen:
https://intersective.atlassian.net/browse/CORE-7535

## changes
- update exist core-based API to graphql for badges list

### DIFF
--- a/projects/v3/src/app/services/achievement.service.ts
+++ b/projects/v3/src/app/services/achievement.service.ts
@@ -1,3 +1,4 @@
+import { ApolloService } from '@v3/services/apollo.service';
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, Observable, of } from 'rxjs';
 import { map, shareReplay } from 'rxjs/operators';
@@ -25,12 +26,17 @@ export interface Achievement {
   image?: string;
   isEarned?: boolean;
   earnedDate?: string;
+  type: string;
+  badge: string;
+  openBadge?: string;
+  progress?: number;
+  active?: boolean;
+  certificateUrl?: string;
 }
 
 @Injectable({
   providedIn: 'root'
 })
-
 export class AchievementService {
   private _achievements$ = new BehaviorSubject<Achievement[]>([]);
   achievements$ = this._achievements$.pipe(shareReplay(1));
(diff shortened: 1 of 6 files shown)


projects/v3/src/app/services/achievement.service.ts
```diff
@@ -39,64 +45,87 @@ export class AchievementService {
...
+    private apolloService: ApolloService,
     private request: RequestService,
     private utils: UtilsService,
     private demo: DemoService,
   ) { }
 
-  getAchievements(order?) {
+  graphQLGetAchievements(): Observable<Achievement[]> {
+    return this.apolloService.graphQLFetch(
+      `query badges {
```
shawnm0705: the `query badges` is not supposed to be used here. It is used for the badge project. We should use `query achievements` instead

projects/v3/src/app/services/achievement.service.ts
```diff
@@ -39,64 +45,126 @@ export class AchievementService {
...
+   * @return  {Observable<Achievement>[]}          achievement list in badges
+   */
+  graphQLGetAchievements(
+    filter?: string,
+    type?: string,
+    active?: boolean
+  ): Observable<Achievement[]> {
+    return this.apolloService
+      .graphQLFetch(
+        `query achievements($filter: String, $type: String, $active: Boolean) {
```
trtshen: @shawnm0705 now using `query achievements`

shawnm0705: When we use `query achievements` in the APP, we don't need to worry about the arguments of this endpoint. They are all built for LearnFlow. We can just query it without any arguments

projects/v3/src/app/services/achievement.service.ts
```diff
@@ -39,64 +45,126 @@ export class AchievementService {
...
+        }
+      )
+      .pipe(
+        map(
+          (res: {
+            data: {
+              badges: Achievement[];
+            };
+          }) => {
+            return res?.data?.badges || [];
```
shawnm0705: the response of this endpoint will be `res.data.achievements`. Please merge this code to trunk and test on stage first, before merging it to the golive branch

shawnm0705 (changes requested):

trtshen:
@shawnm0705 https://github.com/intersective/app/pull/2271#discussion_r1804287418
i did what you said, but the validation didnt get through.

Screenshot 2024-10-17 at 4 04 03 pm

and i think it's fine to just keep it as it wont hurt and can be future proof too.

trtshen:
done, @shawnm0705 please help review again. it;s ok on staging now.
Screenshot 2024-10-17 at 4 33 03 pm

shawnm0705 (approved):

PR #2526, opened 2025-08-26
[CORE-7995] added clear all button for all stale unlock indicator

trtshen:
- mark all indicator as read (clear all stale one)


trtshen:
Closing note: no longer needed, team decided this is not necessary as the underlying problem is todolist from api

PR #2529, opened 2025-08-28
[CORE-7499] slider type question UI - trunk

trtshen:
- implemented slider question type UI & logic


trtshen:
Closing note: This PR was branched out from trunk, shouldn't be use for golive PR

PR #2608, opened 2026-02-04
[CORE-8140] 2.4.y.z/missing-pulsecheck

trtshen:


trtshen:
Review rating is part of the process before pulsecheck can be displayed on the app.
Backend logic (by design): Review rating submission -> pulsecheck popup (in case reviewrating isn't submitted, pulsecheck is also skipped)

